PR #109, opened 2015-01-12
Added the missing package for abrt

tkolhar:
This step is missing :
https://github.com/theforeman/foreman_abrt/blob/katello-docs/README.md#installing-the-foreman-plugin

Added as per the instructions given in README
please review it thanks

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -393,12 +393,15 @@ def setup_abrt():
     packages = [
         'abrt-cli',
         'rubygem-smart_proxy_abrt',
-        'rubygem-smart_proxy_pulp'
+        'rubygem-smart_proxy_pulp',
+        'ruby193-rubygem-foreman_abrt'
     ]
     for package in packages:
         run('yum install -y {0}'.format(package))
 
     run('systemctl restart foreman')
+    # workaround as sometimes foreman service does not restart with systemctl
+    run('touch /usr/share/foreman/tmp/restart.txt')
 
     # edit the config files
     host = env['host']


automation_tools/__init__.py
```diff
@@ -393,12 +393,14 @@ def setup_abrt():
...
         'rubygem-smart_proxy_abrt',
-        'rubygem-smart_proxy_pulp'
+        'rubygem-smart_proxy_pulp',
+        'ruby193-rubygem-foreman_abrt'
     ]
     for package in packages:
         run('yum install -y {0}'.format(package))
 
     run('systemctl restart foreman')
+    run('touch /usr/share/foreman/tmp/restart.txt')
```
sghai: ACK

sghai:
ACK

kbidarkar:
ACK. Will leave it as it is for elyezer to merge.

elyezer:
ACK

automation_tools/__init__.py
```diff
@@ -393,12 +393,14 @@ def setup_abrt():
...
```
elyezer: The above will restart foreman, is there anything else touching the restart file does?

elyezer:
ACK, pending comment.

automation_tools/__init__.py
tkolhar: @elyezer  it only restarts the foreman as per discussion with developer

sghai: `touch` is either to create empty file or to change the timestamps of file..forman is already restarting at line 402

elyezer: @tkolhar then I think is better to remove this line as @sghai said, foreman is already restarting at line 402.

tkolhar: @sghai @elyezer  :  I had a talk with the developer
this was work around provided to me in case foreman service does
not gets restarted with systemctl

omaciel:
@tkolhar @elyezer @sghai @kbidarkar one minor but important issue here is that, if we add a workaround to our code, we have to make sure that:

* There a BZ associated with the issue
* We don't use the workaround during the release cycle

elyezer:
@omaciel good point.

PR #111, opened 2015-01-13
Setup default docker on product_install task

elyezer:
Docker should be available to be used as a compute resource. Because
this setting up the default docker is needed to allow testing docker
product features.


Ichimonji10:
ACK

JacobCallahan:
ACK

PR #112, opened 2015-01-14
Improve errata tasks

elyezer:
Make manage_repos disable beaker repos instead of removing them.

Enable beaker repos when running errata_upgrade task.

Also drop a not required environment variable from errata_upgrade
docstring.


omaciel:
APT

Ichimonji10:
APT

PR #113, opened 2015-01-14
Enable required repository for setting up docker

elyezer:
To install docker is needed to enable the
rhel-<os_version>-server-extras-rpm repository. Enable it when setting
up default docker.


Ichimonji10:
APT

omaciel:
APT x 2

PR #114, opened 2015-01-15
Simplify commands

Ichimonji10:
Drop unnecessary quotes. The following four commands:

    $ echo other_args=foo
    $ echo other_args="foo"
    $ echo other_args=\"foo\"
    $ echo "other_args=\"foo\""

result in this output:

    other_args=foo
    other_args=foo
    other_args="foo"
    other_args="foo"

The outermost quotes serve no function beyond making the command harder to read.

Use raw strings. The following two strings are equivalent:

    'before \\ after'
    r'before \ after'

Make some other changes that make the code cleaner.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -251,18 +251,17 @@ def setup_default_docker():
 
     # SElinux workaround let us use ``http://localhost:2375`` for a
     # ``Docker`` Compute Resurce.
-    options = [
+    options = ' '.join([
         '--selinux-enabled',
         '-H tcp://0.0.0.0:2375',
         '-H unix:///var/run/docker.sock',
-    ]
-
+    ])
     if os_version >= 7:
-        run('echo "OPTIONS={0}" >> /etc/sysconfig/docker'
-            ''.format(' '.join(options)))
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I think this will fix the RHEL 6 provisioning issue in Jenkins. Even if it doesn't, the code in question benefits from this change due to improved legibility.

elyezer:
ACK

Ichimonji10:
Here's the relevant Jenkins error:

    run: echo "other_args=\"--selinux-enabled -H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock\"" >> /etc/sysconfig/docker
    out: -H: -c: line 0: unexpected EOF while looking for matching `"'
    out: -H: -c: line 1: syntax error: unexpected end of file

Ichimonji10:
Code has been tested on a live machine to make sure that it works. It gets to [`docker ps`](https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L274) before failing.

elyezer:
ACK

JacobCallahan:
ACK

PR #115, opened 2015-01-16
Fix command which populates /etc/sysconfig/docker

Ichimonji10:
The `setup_default_docker` fabric task is currently broken. It populates
`/etc/sysconfig/docker` on RHEL 6 remote hosts with this line:

    other_args=

Fix this so that the configuration file is populated with a proper value:

    other_args='--selinux-enabled true --host tcp://0.0.0.0:2375 --host unix:///var/run/docker.sock'

The relevant command is tricky because Python, fabric and bash each have a hand
in transforming the string before it is executed.


elyezer:
APT

JacobCallahan:
ACK

PR #116, opened 2015-01-16
Fix task `setup_default_docker`

Ichimonji10:
The `docker` command-line tool requires that an equals sign be used with the
`--selinux-enabled` option. This is valid:

    --selinux-enabled=true

This is invalid:

    --selinux-enabled true


elyezer:
APT

JacobCallahan:
ACK

PR #120, opened 2015-01-21
Install `docker-io` from `epel` if using RHEL 6.

### LINKED ISSUE #119: Install epel repo when setting up default docker for RHEL 6
Docker package is only available on epel repo for RHEL 6. The setup default capsule should install/enable epel repository before installing `docker-io` package.

omaciel:
We need to enable the `epel` repository to install `Docker` on RHEL 6.
Fix #119.


elyezer:
APT

Ichimonji10:
APT

PR #122, opened 2015-01-23
fixed the pattern for abrt setup

tkolhar:
please review it
thanks

 cat file.txt
#SSLVerify = no
#SSLClientAuth =

 sed -i -e "s|#SSLVerify = no.*|SSLVerify = yes|" file.txt
 cat file.txt
SSLVerify = yes
#SSLClientAuth =

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -393,12 +393,15 @@ def setup_abrt():
     packages = [
         'abrt-cli',
         'rubygem-smart_proxy_abrt',
-        'rubygem-smart_proxy_pulp'
+        'rubygem-smart_proxy_pulp',
+        'ruby193-rubygem-foreman_abrt'
     ]
     for package in packages:
         run('yum install -y {0}'.format(package))
 
     run('systemctl restart foreman')
+    # workaround as sometimes foreman service does not restart with systemctl
+    run('touch /usr/share/foreman/tmp/restart.txt')
 
     # edit the config files
     host = env['host']


Ichimonji10:
Can you provide some context for this PR, @tkolhar? I well understand the regular expression. But what will this change impact? Is the `setup_abrt` task used when provisioning upstream? Downstream? Does `setup_abrt` not work right now? Can you give an example of what the configuration file looks like? The change itself looks sane, but I would like some contextual information here.

tkolhar:
@Ichimonji10 , the pattern s/^SSLClientAuth = .*/SSLClientAuth = puppet/ is not working in upstream
so i had to modify the pattern s|#SSLClientAuth = .*|SSLClientAuth = puppet| . requesting you
to please let me know if it is clear know . thanks.

elyezer:
This will fail too because there is a space after `#`, grabbed this from the config file:

```conf
# no means that ssl certificates will not be checked
# SSLVerify = no

...

# Using Puppet certificate:
# SSLClientAuth = puppet
```

elyezer:
Tried manually:

```shell
$ sed -e "s|#SSLVerify = no.*|SSLVerify = yes|" /etc/libreport/plugins/ureport.conf
...

# no means that ssl certificates will not be checked
# SSLVerify = no

...
```

With the space:

```shell
$ sed -e "s|# SSLVerify = no.*|SSLVerify = yes|" /etc/libreport/plugins/ureport.conf
...

# no means that ssl certificates will not be checked
SSLVerify = yes

...
```

Also in this case using `/` or `|` as a separator does not make any difference. Be aware that I have removed the `-i` parameter from my example because that flag, when present, edit the file in place and does not output anything.

PR #125, opened 2015-01-28
Update iso_install task

elyezer:
ISO URL is now being read from environment variable. Also accepts the
ISO_URL or BASE_URL as the environment variable name.

Make iso_install accept the admin_password as parameter, and drop
iso_url parameter in favor of the environment variable.

Finally, improve the task's docstring.


sthirugn:
ACK

omaciel:
ACK

PR #127, opened 2015-01-29
Remove default org and loc from upstream_install

elyezer:
Upstream and downstream installer now defines default organization and
location following the same name. Update upstream_install task to not
specify initial-organization and initial-location.

From the katello-installer help:

    --foreman-initial-location    Name of an initial location (default:
                                  "Default Location")
    --foreman-initial-organization  Name of an initial organization
                                    (default: "Default Organization")


omaciel:
ACK and roll

PR #128, opened 2015-01-29
Improve iso_download task

elyezer:
Append / if the url does not ends with .iso and does not have it in
other to urljoin not drop the last section.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1065,6 +1065,10 @@ def iso_download(iso_url=None):
         # server
         iso_filename = None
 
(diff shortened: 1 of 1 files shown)


sthirugn:
ACK, but not sure why we need to support directory path here.  Cant we just make it a requirement to supply the direct .iso link?

omaciel:
ACK and merge (the explanation will be provided as a comment but it won't block the merge)

elyezer:
@sthirugn by accepting the directory path a generic URL can be provided for the trigger job and we are able to generate URL for the iso and downstream jobs.

sthirugn:
@elyezer ty

PR #129, opened 2015-01-29
Setup ABRT on RHEL 6 machines

elyezer:
Now ABRT could be set up on RHEL 6 machines


omaciel:
ACK (context: we want to install ``ABRT`` regardless of the version for RHEL.)

cswiii:
ACK

PR #130, opened 2015-02-02
Improve setup_abrt task

elyezer:
Make setup_abrt distinguish which OS version it is running against in
order to run `service` for RHEL 6, and `systemctl` for RHEL 7+.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -416,7 +416,12 @@ def setup_abrt():
     for package in packages:
         run('yum install -y {0}'.format(package))
 
-    run('systemctl restart foreman')
+    os_version = distro_info()[1]
+
+    if os_version >= 7:
+        run('systemctl restart foreman')
+    else:
+        run('service foreman restart')
     # workaround as sometimes foreman service does not restart with systemctl
     run('touch /usr/share/foreman/tmp/restart.txt')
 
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -428,8 +433,12 @@ def setup_abrt():
...
-    run('systemctl start abrt-ccpp')
+    if os_version >= 7:
+        run('systemctl start abrtd')
+        run('systemctl start abrt-ccpp')
+    else:
+        run('service abrtd start')
+        run('service abrt-ccpp start')
 
     # edit the config files
     run('sed -i -e "s|^URL = .*|URL = https://{0}:8443/abrt/|" '
```
elyezer: If you will test the task be aware that the file sed edits is not in place for RHEL 6. I still investigating/researching to find why.

automation_tools/__init__.py
```diff
@@ -428,8 +433,12 @@ def setup_abrt():
         '/etc/foreman-proxy/settings.d/abrt.yml')
 
     # run the required commands
-    run('systemctl start abrtd')
-    run('systemctl start abrt-ccpp')
+    if os_version >= 7:
+        run('systemctl start abrtd')
+        run('systemctl start abrt-ccpp')
```
Ichimonji10: FYI, you can place all of the units being started on one line:

    systemctl start abrtd abrt-ccpp

I'm not terribly worried about making this change right now, though. As per a conversation with @cswiii, we may move SysV/systemd daemon management logic into a separate function. Something like this:

```python
def manage_daemon(action, daemon):
    if distro_info()[1] == 6:
        command = 'service {} {}'.format(daemon, action)
    else:
        command = 'systemctl {} {}'.format(action, daemon)
    return run(command)
```

Ichimonji10:
ACK

cswiii:
ACK, recommending we move all sysv/systemd logic to its own separate function.

https://github.com/SatelliteQE/automation-tools/issues/131

automation_tools/__init__.py
sthirugn: +1 to this suggestion

PR #132, opened 2015-02-03
Add manage_daemon helper function

### LINKED ISSUE #131: Put RHEL version logic into separate service() type function
Now that we are/will be supporting various service calls, it would probably be better to isolate the logic on whether to use sysV's 'service' as seen in RHEL6 or systemd's  'sysctl' as seen in RHEL7, into its own function. 

Doing this would:

a) help avoid faulty logic and/or duplicitous code every time a new service needs to be executed 
b) help assure that there is automation coverage against both RHEL versions rather than inadvertently missing one.

Logic might be something like this pseudocode

<pre>
def os_service(daemon, state):
    rhelver = determine_distro()
    if rhelver = "rhel7":
        exec(sysctl state daemon)
    if rhelver = "rhel6":
        exec(service daemon state)
</pre>

Thus, subsequent automation could be triggered easily and mostly hassle free, regardless as to distro.

<pre>
os_service(abrtd, start)
os_service(squid, restart)
</pre>

elyezer:
This function helps handle daemon actions and will run corresponding
SysV or systemd command based on the OS version.

Closes #131

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -131,10 +131,8 @@ def setup_proxy():
 
     Proxy information is passed using the PROXY_INFO environmental variable.
     e.g. PROXY_INFO=proxy://root:myP4$$@myhost.awesomedomain.com:8888
+
     """
-
-    os_version = distro_info()[1]
-
     proxy_info = urlsplit(os.environ.get('PROXY_INFO'))
     if not proxy_info.hostname or not proxy_info.port:
         raise Exception("You must include the proxy hostname and port.")
@@ -160,15 +158,15 @@ def setup_proxy():
         ))
 
     # restart the capsule-related services
-    daemons = (
-        'httpd', 'pulp_celerybeat', 'pulp_resource_manager', 'pulp_workers',
-    )
-    if os_version >= 7:
-        run('systemctl restart {0}'.format(' '.join(daemons)), warn_only=True)
+    if distro_info()[1] >= 7:
         run('yum install -y iptables-services')  # Install firewall package
-    else:
-        for daemon in daemons:
-            run('service {0} restart'.format(daemon), warn_only=True)
+    for daemon in (
+            'httpd',
+            'pulp_celerybeat',
+            'pulp_resource_manager',
+            'pulp_workers',
+    ):
+        manage_daemon('restart', daemon, warn_only=True)
 
     # Satellite 6 IP
     sat_ip = search(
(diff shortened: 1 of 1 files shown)


Ichimonji10:
I've manually tested this and can confirm that the following succeeds when run against a RHEL 6.6 server:

    >>> from automation_tools import manage_daemon
    >>> manage_daemon('status', 'iptables')

However, that same code fails when run against a RHEL 7.0 server. Can anyone else reproduce?

Ichimonji10:
Oh, by the way, this is the error I get:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "automation_tools/__init__.py", line 1469, in manage_daemon
    return run(command, pty=pty, warn_only=warn_only)
  File "/home/ichimonji10/.virtualenvs/automation-tools/lib/python2.7/site-packages/fabric/network.py", line 647, in host_prompting_wrapper
    return func(*args, **kwargs)
  File "/home/ichimonji10/.virtualenvs/automation-tools/lib/python2.7/site-packages/fabric/operations.py", line 1048, in run
    shell_escape=shell_escape)
  File "/home/ichimonji10/.virtualenvs/automation-tools/lib/python2.7/site-packages/fabric/operations.py", line 901, in _run_command
    _prefix_commands(_prefix_env_vars(command), 'remote'),
  File "/home/ichimonji10/.virtualenvs/automation-tools/lib/python2.7/site-packages/fabric/operations.py", line 710, in _prefix_env_vars
    return shell_env_str + command
TypeError: cannot concatenate 'str' and 'tuple' objects
```

automation_tools/__init__.py
```diff
@@ -1514,3 +1441,29 @@ def update_packages(*args, **kwargs):
...
+
+    :param str action: Daemon action like start, stop, restart, info
+    :param str daemon: Daemon name to perform the action
+    :param bool pty: Controls the creation of a pseudo-terminal when managing
+        the daemon. Some daemons actions fail with pty=True.
+    :param bool warn_only: Will be passed directly to Fabric's run
+
+    """
+    if distro_info()[1] >= 7:
+        command = 'systemctl {} {}'.format(action, daemon)
```
elyezer: @Ichimonji10 was a trailing `,` in this line. Have copied more than needed.

Thanks for catching this

Ichimonji10:
The command now succeeds when run against RHEL 66 or RHEL 7. Thank you.

ACK

omaciel:
ACK

PR #135, opened 2015-02-04
Improve manage_repos task

elyezer:
Now manage_repos figure out the distro version instead of expecting it
to be passed as a parameter.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -683,12 +683,8 @@ def install_prerequisites():
     run('iptables-save > /etc/sysconfig/iptables')
 
 
-def manage_repos(os_version=None, cdn=False):
+def manage_repos(cdn=False):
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -817,10 +812,8 @@ def cdn_install():
     """
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
-    os_version = distro_info()[1]
-
     # Enable some repos
-    manage_repos(os_version, True)
+    manage_repos(True)
```
omaciel: Maybe explicitly pass `cdn=True` here?

omaciel:
ACK pending comment.

elyezer:
Comments addressed

Ichimonji10:
ACK

Issue #133, opened 2015-02-04
Separate task for enabling/disabling SELinux


omaciel:
We have a total of 4 places in the code where we handle turning SELinux ON/OFF. We should create a specific task for this and update the code to use it, as it would simplify editing this feature's default behavior.

PR #136, opened 2015-02-04
Add setenforce helper function

### LINKED ISSUE #134: SELinux should be a flag for SELinux task
Assuming we implement #133 then I would like to suggest that we make the value for the `SELinux` mode to be an optional argument that can be passed from the command line. If no value is passed then we should default to `Enforcing`.

elyezer:
This function modifies the mode SELinux is running in.

Install tasks don't set the SELinux mode in favor of being more atomic
and just dealing more with the installation steps.

Call setenforce just before the install on product_install task. The
enforce mode used, will be the first found from function parameter or
environment variable or the default enforcing mode.

I've created two commits to handle #133 and #134.


omaciel:
ACK

Ichimonji10:
ACK

PR #138, opened 2015-02-05
Run setup_abrt in warn_only mode on RHEL 6

elyezer:
Setting up ABRT on RHEL 6 machines is failing and running in warn_only
mode it will still run the task but will not fail product_install
because the setup failure. And, when setup ABRT start working on RHEL 6,
the task will work properly.


sghai:
ACK

omaciel:
ACK

PR #140, opened 2015-02-05
Fix execution of setenforce task

elyezer:
setenforce is now being run by the execute function but its parameter
was not handle properly. This fixes that problem and properly call
setenforce using execute function.


Ichimonji10:
APT

omaciel:
ACK

PR #141, opened 2015-02-05
Fix distro_info execution on product_install task

elyezer:
As product install can run the tasks on the specified server on in a
virtual machine created, because this tasks should be executed using the
execute function. Execute distro_info on the correct host.


elyezer:
This kind of mistake should not happen in the future, have been a long time without dealing with product install execution model.

JacobCallahan:
APT

Ichimonji10:
ACK

Ichimonji10:
For reference:

> \<ichimonji10> elyezer: The dict returned only has one item?
> \<elyezer> ichimonji10: in that case yes

PR #142, opened 2015-02-05
Add sam_upstream_install task

elyezer:
This task will install SAM upstream using the katello-deploy script.
Also add sam-upstream as a valid distribution for product_install task.


sthirugn:
ACK

Ichimonji10:
ACK

PR #143, opened 2015-02-05
Improve setup_vm_provisioning task

elyezer:
Avahi is needed in order to discover the virtual machines IP addresses.
As being a requirement, setup avahi when runnin setup_vm_provisioning
task.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -567,6 +567,7 @@ def setup_vm_provisioning(interface=None):
 
     # Install other required packages
     packages = (
+        'avahi',
         'bash',
         'bridge-utils',
         'cloud-utils',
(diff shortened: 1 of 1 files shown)


omaciel:
APT

automation_tools/__init__.py
```diff
@@ -584,6 +585,10 @@ def setup_vm_provisioning(interface=None):
     )
     run('yum install -y {0}'.format(' '.join(packages)))
 
+    # Setup avahi
+    manage_daemon('start', 'avahi-daemon')
+    manage_daemon('enable', 'avahi-daemon')
```
Ichimonji10: I was surprised that the systemd unit file (and, presumably, the SysV init script) is named avahi-daemon, not avahid. But I quickly logged in to my home server, and on that machine at least, avahi-daemon is the correct name. I've been schooled. At least it's a lot more sane than `NetworkManager`. :p

Ichimonji10:
ACK

automation_tools/__init__.py
elyezer: :smile:

PR #144, opened 2015-02-05
Improve product_install for sam installation

elyezer:
Avoid running unnecessary tasks on SAM product installation.

Also add a satellite6 prefix to Satellite related distributions to make
easier to manage jenkins jobs for both SAM and Satellite.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -892,7 +892,8 @@ def product_install(distribution, create_vm=False, certificate_url=None,
                     selinux_mode=None):
     """Task which install every product distribution.
 
-    Product distributions are cdn, downstream, iso or upstream.
+    Product distributions are sam-upstream, satellite6-cdn,
+    satellite6-downstream, satellite6-iso or satellite6-upstream.
 
     If ``create_vm`` is True then ``vm_destroy`` and ``vm_create`` tasks will
     be run. Make sure to set the required environment variables for those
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -914,11 +914,11 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
     install_tasks = {
-        'cdn': cdn_install,
-        'downstream': downstream_install,
-        'iso': iso_install,
-        'upstream': upstream_install,
         'sam-upstream': sam_upstream_install,
+        'satellite6-cdn': cdn_install,
+        'satellite6-downstream': downstream_install,
+        'satellite6-iso': iso_install,
+        'satellite6-upstream': upstream_install,
```
Ichimonji10: Can you update the documentation to reflect this change?

elyezer: Definitely. Good catch, thank you

automation_tools/__init__.py
```diff
@@ -958,17 +958,19 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+    if distribution in (
+            'satellite6-cdn', 'satellite6-downstream', 'satellite6-iso'):
         execute(setup_default_capsule, host=host)
 
-    execute(setup_default_docker, host=host)
-    # execute returns a dict, the result is the first value. OS version is the
-    # resulting index 1
-    if execute(distro_info, host=host).values()[0][1] == 6:
-        with warn_only():
+    if not distribution.startswith('sam'):
```
Ichimonji10: Would it be more straightforward to get rid of the negative and write this instead?

    if distribution.startswith('satellite6'):

elyezer:
@Ichimonji10 have addressed your comment

Ichimonji10:
ACK pending comments

automation_tools/__init__.py

elyezer:
@Ichimonji10 addressed your comments. Thank you

automation_tools/__init__.py
Ichimonji10: I'm thinking of the docs that are pushed to readthedocs, too.

Ichimonji10: I say this because I rely heavily on those when setting up systems using automation-tools. I'm your audience of one. :)

sthirugn:
ACK

automation_tools/__init__.py
elyezer: There we go, yeah I have forgotten about it :disappointed:

elyezer:
Now comments was addressed

PR #145, opened 2015-02-10
Fix a bug in task `run_errata`

Ichimonji10:
Make sure to update settings with the `<package>d-update-settings` command at
the beginning of the `run_errata` task. See:
https://bugzilla.redhat.com/show_bug.cgi?id=1182352

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1411,13 +1411,12 @@ def errata_upgrade():
 
 
 def run_errata():
-    """Run the errata to upgrade packages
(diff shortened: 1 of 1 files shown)


JacobCallahan:
ACK

sthirugn:
To add more context, the actual bug was not in automation-tools and this fix by @Ichimonji10  will be a temporary workaround in automation-tools until the main bug (listed in the first comment) is fixed.

sthirugn:
ACK

PR #149, opened 2015-02-19
Open port 5000 TCP on install_prerequisites

elyezer:
Port 5000 TCP are used to publish the Satellite docker registry for
docker repositories. That port need to be opened otherwise clients will
not be able to pull images from it.


Ichimonji10:
ACK

JacobCallahan:
ACK

PR #151, opened 2015-02-24
Subscribe to <major>Server release

elyezer:
Customers are recommended to update to latest minor version available
before installing Satellite. To allow that, subscribe task is now
updated to subscribe to <major>Server release instead of a specific
<major>.<minor> release.


sthirugn:
APT

omaciel:
ACK

PR #152, opened 2015-02-25
Removed redundant calls to install libvirt.

omaciel:
The package ``libvirt`` should only be installed by the
``setup_default_capsule`` task.


JacobCallahan:
makes sense! APT

Ichimonji10:
ACK

PR #154, opened 2015-03-05
Add setup_firewall task

elyezer:
This task will check for the required firewall rules and will add only
if the required port is not already opened.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -370,6 +370,45 @@ def setup_fake_manifest_certificate(certificate_url=None):
     manage_daemon('restart', 'tomcat6' if distro_info()[1] <= 6 else 'tomcat')
 
 
+def setup_firewall():
+    """Setup firewall rules that Satellite 6 needs to work properly"""
+    ports = (
+        # Port 443 for HTTPS (secure WWW) must be open for incoming
+        # connections.
+        443,
+        # Port 5671 must be open for SSL communication with managed systems.
+        5671,
+        # Port 80 for HTTP (WWW) must be open to download the bootstrap files.
+        80,
+        # Port 8140 must be open for incoming Puppet connections with the
+        # managed systems.
+        8140,
+        # Port 9090 must be open for Foreman Smart Proxy connections with the
+        # managed systems.
+        9090,
+        # Port 22 must be open for connections via ssh
+        22,
+        # Port 5000 must be open for Docker registry communication.
+        5000,
+    )
+
+    for port in ports:
+        rule_exists = run(
+            r'iptables -nL INPUT | grep -E "^ACCEPT\s+tcp.*{0}"'.format(port),
+            quiet=True,
+            warn_only=True
+        ).return_code == 0
+        if not rule_exists:
+            run(
+                'iptables -I INPUT -m state --state NEW -p tcp --dport {0} '
+                '-j ACCEPT'.format(port)
+            )
+
+    # To make the changes persistent across reboots when using the command line
+    # use this command:
+    run('iptables-save > /etc/sysconfig/iptables')
+
+
 def setup_abrt():
     """Task to setup abrt on foreman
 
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -370,6 +370,45 @@ def setup_fake_manifest_certificate(certificate_url=None):
...
+        # managed systems.
+        9090,
+        # Port 22 must be open for connections via ssh
+        22,
+        # Port 5000 must be open for Docker registry communication.
+        5000,
+    )
+
+    for port in ports:
+        have_rule = run(
```
omaciel: s/have_rule/rule_exists ?

automation_tools/__init__.py
```diff
@@ -370,6 +370,45 @@ def setup_fake_manifest_certificate(certificate_url=None):
...
+            r'iptables -nL | grep -E "^ACCEPT\s+tcp.*{0}"'.format(port),
```
elyezer: I'm doing this check because I don't want duplicated rules to be added. The puppet module that setup_default_capsule already add all necessary rules but not for 5000 port. I also want that when just this task is run it creates all necessary rules.

omaciel:
ACK

automation_tools/__init__.py
```diff
@@ -370,6 +370,45 @@ def setup_fake_manifest_certificate(certificate_url=None):
...
+            quiet=True,
+            warn_only=True
+        ).return_code == 0
+        if not have_rule:
```
elyezer: I have created an one use variable because the if was difficult to read, this way it is easier to read

automation_tools/__init__.py
elyezer: thanks will change

automation_tools/__init__.py
```diff
@@ -370,6 +370,45 @@ def setup_fake_manifest_certificate(certificate_url=None):
...
+        9090,
+        # Port 22 must be open for connections via ssh
+        22,
+        # Port 5000 must be open for Docker registry communication.
+        5000,
+    )
+
+    for port in ports:
+        rule_exists = run(
+            r'iptables -nL | grep -E "^ACCEPT\s+tcp.*{0}"'.format(port),
```
Ichimonji10: It wouldn't hurt to specify that you're working with the INPUT iptables chain, like so:

    iptables -nL INPUT

This is because there are also FORWARD and OUTPUT tables, and you want to ignore them.

Also, remember that GNU grep has a very unusual syntax for specifying character classes. Whereas perl and many other regex engines specify the "whitespace" character class with `\s`, GNU grep uses `[[:space:]]`. As a result, the regex you've written should fail.

In addition, I recommend adding `$` to the end of the regex, as it is possible that an IP address could match what you're searching for:

Ichimonji10: It wouldn't hurt to specify that you're working with the INPUT iptables chain, like so:

    iptables -nL INPUT

This is because there are also FORWARD and OUTPUT tables, and you want to ignore them.

Also, remember that GNU grep has a very unusual syntax for specifying character classes. Whereas perl and many other regex engines specify the "whitespace" character class with `\s`, GNU grep uses `[[:space:]]`. As a result, the regex you've written should fail. This expression succeeds on my home server:

    "^ACCEPT[[:space:]]\+tcp.*443"

In addition, I recommend adding `$` to the end of the regex, as it is possible that an IP address could match what you're searching for:

    ACCEPT     tcp  --  0.0.0.0/0            192.168.0.80/0            tcp dpt:443

Finally, in GNU grep, there is no difference between basic and extended regular expressions, so the `-E` flag can be dropped. If the suggestions above are taken, then the command becomes:

    r'iptables -nL INPUT | grep "^ACCEPT[[:space:]]\+tcp.*{0}$"'.format(port),

I've checked the above against `man grep` on several of our RHEL systems. I've also taken the firewall output from my home server, SCP'd it to one of our RHEL systems, and ran grep against the file. Everything checks out. (There are no firewall rules on the RHEL systems I have right now, and I don't want to mess things up by adding some.)

Ichimonji10: While we're at it, I should mention that `iptables` has facilities for checking for firewall rules. For example, the `--check` option looks promising.

Ichimonji10: Again, from my home server:

    # iptables-save | grep 22
    -A INPUT -p tcp -m tcp --dport 22 -j ACCEPT
    # iptables --check INPUT -p tcp -m tcp --dport 22 -j ACCEPT; echo $?
    0
    # iptables --check INPUT -p tcp -m tcp --dport 23 -j ACCEPT; echo $?
    iptables: Bad rule (does a matching rule exist in that chain?).
   1 
    # iptables --check INPUT -p tcp -m tcp --dport 22; echo $?
    iptables: Bad rule (does a matching rule exist in that chain?).
    1

elyezer: I'm using -E with allows me to use the extended regex support :smile: 

Will change to look for the INPUT chain. And the regex does not end with $ because rules added by the puppet module is like this `ACCEPT     tcp  --  0.0.0.0/0            0.0.0.0/0           multiport ports 80,443 /* 100 accept http/https */`

About `--check` iptables option will try to improve using that later.

Ichimonji10: `-E` makes a difference on a mac?

Ichimonji10:
ACK

PR #155, opened 2015-03-06
Update setup_default_docker task

elyezer:
Recent installations of Docker creates the dockerroot user and group
instead of docker. Update the task to check for both and add the
available group to the apache user.


omaciel:
ACK

PR #156, opened 2015-03-06
Drop warn_only when quite is True

elyezer:
From the fabric.operations.run documentation:

    To both ignore non-zero return codes and force a command to run
    silently, specify quiet=True.


omaciel:
ACK

Ichimonji10:
ACK

PR #157, opened 2015-03-06
Check for failed and succeeded on run result

elyezer:
From the fabric.operations.run documentation:

    run will return the result of the remote program’s stdout as a
    single (likely multiline) string. This string will exhibit failed
    and succeeded boolean attributes specifying whether the command
    failed or succeeded, and will also include the return code as the
    return_code attribute.


omaciel:
ACK

elyezer:
Update: rebased

Ichimonji10:
ACK

PR #160, opened 2015-03-12
Task to configure system for email notification.

omaciel:
This task will configure your Satellite 6 system to send out email
notifications. The task itself needs to restart the entire stack in
order to take effect (via katello-service restart). I chose not to
restart the stack within the task itself, but to let the user call it
when appropriate, preferably before another task that may already
restart all services, such as the ``setup_default_capsule`` task.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -353,6 +353,23 @@ def setup_default_capsule(interface=None):
     )
 
 
+def setup_email_notification(smtp=None):
+    """Configures system to handle email notification.
+
+    NOTE: this task needs a 'katello-service restart', so choose wisely when to
+    call it, preferably before another task that restarts the stack, such as
+    the 'setup_default_capsule' task.
+
+    :param str smtp: A valid URL to a SMTP server.
+
+    """
+
+    # edit the config file
+    if smtp is not None:
+        run('sed -i -e "s|address.*|address: {0}|" '
+            '/etc/foreman/email.yaml'.format(smtp))
+
+
 def setup_fake_manifest_certificate(certificate_url=None):
     """Task to setup a fake manifest certificate
 
--- a/fabfile.py
+++ b/fabfile.py
@@ -25,6 +25,7 @@ from automation_tools import (
     setup_ddns,
     setup_default_capsule,
     setup_default_docker,
+    setup_email_notification,
     setup_fake_manifest_certificate,
     setup_firewall,
     setup_proxy,


automation_tools/__init__.py
```diff
@@ -353,6 +353,28 @@ def setup_default_capsule(interface=None):
...
+    """
+
+    # edit the config files
+    if smtp is not None:
+        run('sed -i -e "s|address.*|address: {0}|" '
+            '/etc/foreman/email.yaml'.format(smtp))
+
+        hostname = run('hostname', quiet=True).strip()
+        if hostname:
+            domain = hostname.split('.', 1)[1]
```
sthirugn: I am not sure about the purpose of the domain value.  Can you please test and make sure that this actually works?

automation_tools/__init__.py
```diff
@@ -353,6 +353,28 @@ def setup_default_capsule(interface=None):
...
+def setup_email_notification(smtp=None):
+    """Configures system to handle email notification.
+
+    NOTE: this task needs a 'katello-service restart', so choose wisely when to
+    call it, preferably before another task that restarts the stack, such as
+    the 'setup_default_capsule' task.
+
+    """
+
+    # edit the config files
```
sthirugn: s/files/file

automation_tools/__init__.py
```diff
@@ -353,6 +353,28 @@ def setup_default_capsule(interface=None):
     )
 
 
+def setup_email_notification(smtp=None):
```
elyezer: This task seems to just configure the local email service. When the smtp has a value it will contain an URL with username, password, server and port? Maybe some doc about this parameter would make the task's docstring better.

elyezer:
ACK, pending comment

automation_tools/__init__.py
omaciel: @elyezer the smtp param would only be an url (there's a separate field in the email.yaml file for authentication), but I'll document it.

automation_tools/__init__.py
omaciel: Tested and it didn't seem to affect the end result. I still managed to get notifications. If anything, the default for ``domain`` is something like ``example.com`` which my proposed code would change.

automation_tools/__init__.py
elyezer: Thanks

automation_tools/__init__.py
```diff
@@ -353,6 +353,30 @@ def setup_default_capsule(interface=None):
...
+    """Configures system to handle email notification.
+
+    NOTE: this task needs a 'katello-service restart', so choose wisely when to
+    call it, preferably before another task that restarts the stack, such as
+    the 'setup_default_capsule' task.
+
+    :param str smtp: A valid URL to a SMTP server.
```

elyezer:
ACK

automation_tools/__init__.py
omaciel: I think I want to remove the code that updates the domain attribute as I don't think it buy us anything

sthirugn: +1

omaciel:
I decided to drop the code that updated the ``domain`` attribute as it didn't seem to really affect this feature.

sthirugn:
ACK

elyezer:
Time to merge

elyezer:
Karma given

elyezer:
I can haz beers? `omaciel has 205 points`

PR #162, opened 2015-03-16
Update firewall ports to be opened

elyezer:
According to [1] the setup_firewall task is missing opening some ports,
update the task to open all required ports.

[1] http://www.katello.org/docs/user_guide/capsules/isolation.png


omaciel:
ACK

Ichimonji10:
APT

PR #163, opened 2015-03-17
Improve setup_proxy task

elyezer:
Allow the setup_proxy setup a non-authenticated proxy. To do that, the
username and password expected in the proxy URL can be omitted. Also run
it on product_install if the PROXY_INFO environment variable is set.


Ichimonji10:
APT

omaciel:
ACK

omaciel:
```log
17:55:07 omaciel: elyezer++
17:55:08 robottelo: elyezer has 482 points
```

elyezer:
@omaciel thanks

PR #164, opened 2015-03-18
Fix a typo on firewall ports to be open

elyezer:
Was a typo in the docs where pointed to open ports for qpidd, update the
firewall ports to the real ports. Thanks @omaciel who persisted on this
issue until finding the typo.


JacobCallahan:
APT

Ichimonji10:
APT

PR #165, opened 2015-03-20
Refactor product installer

elyezer:
Refactor the product_install task to avoid calling the katello-installer
several times. All katello-installer options will be aggregated to run
the katello-installer command at the very end, just before setting the
fake manifest.

Add an argument to the tasks which calls katello-installer to define if
the katello-installer should be run or not. If not, then the task will
return the options it should be using on the katello-installer.

Clean the setup_proxy task, now just the options on the
katello-installer are enough.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -13,7 +13,7 @@ import time
 from re import search
 from urlparse import urlsplit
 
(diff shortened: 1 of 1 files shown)


omaciel:
ACK and merge (I shoulder surfed this entire PR and we manually tested it).

PR #166, opened 2015-03-23
Fix setup_proxy task

elyezer:
In the previous refactor was checking for proxy username and password
being None to add the installer options. That is not right as we want to
set the username and password if they are not None.

All credits should go to @sghai who got this issue.


omaciel:
APT

sghai:
ACK.. thanks elyezer for quick fix..

PR #167, opened 2015-04-01
Update the command to get capsule forwarders

elyezer:
Update the regex used by grep in order to grab only valid lines when
reading the /etc/resolv.conf file in setup_default_capsule task.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -229,8 +229,10 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
     if isinstance(run_katello_installer, str):
         run_katello_installer = (run_katello_installer.lower() == 'true')
 
-    forwarders = run('cat /etc/resolv.conf | grep nameserver | '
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -229,8 +229,10 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
     if isinstance(run_katello_installer, str):
         run_katello_installer = (run_katello_installer.lower() == 'true')
 
-    forwarders = run('cat /etc/resolv.conf | grep nameserver | '
-                     'awk \'{print $2}\'', quiet=True).split('\n')
+    forwarders = run(
+        'cat /etc/resolv.conf | grep "^nameserver" | awk \'{print $2}\'',
```
elyezer: Basically the change was `s/grep nameserver/grep "^nameserver"`. Other changes was for better formatting.

omaciel:
ACK

cswiii:
ACK

PR #168, opened 2015-04-07
Don't run manage repos inside a install task

elyezer:
In order to make automation-tools tasks more atomic, don't run
manage_repos inside a install task.

Also make product_install run it before running a install task.


omaciel:
APT

Ichimonji10:
ACK

Issue #72, opened 2014-11-19
Provide ability to create new repofile after installing 


cswiii:
Let's say we want to test an upgrade after installing a product (say, GA compose).  The cycle might be something like

* Install product
* Run populate test
* Point server to repo with latest compose
* Install new updates and restart

If we get issues #71 resolved, it seems it could work something like

* cdn_install
* (automated population)
* generate_compose_repofile
* update_rpms

This would help catch dependency issues in rpm or the like, as well.

elyezer:
#170 is merged, closing this.

PR #170, opened 2015-04-15
Refactor repository management

elyezer:
Create tasks to enable and disable repositories. Also drop manage_repos
in favor of enable_repos task. The Satellite repo utilities was moved to
the enable_satellite_repos task and disabling beaker repos was moved
into its own task, the disable_beaker_repos task.

Also add tasks to work with custom repositories files. The new tasks are
create_custom_repos, delete_custom_repos and manage_custom_repos. The
first will create custom repositories files and the second will remove
them. The latter can enable or disable a custom repo depending on the
boolean value specified for the matching repository name.

Add two modules: repository and utils. This is the first step on
organizing better the code base. The utils module will have utilities
tasks to help creating tasks like distro_info and update_packages tasks.
On the other hand repository module have tasks specific to managing
repositories.

With the new repositories tasks #72 can be closed.

### DIFF
--- a/automation_tools/repository.py
+++ b/automation_tools/repository.py
@@ -1,0 +1,157 @@
+"""Taks for managing repositories"""
+from __future__ import print_function
+
+import sys
+
+from automation_tools.utils import distro_info
+from fabric.api import put, run
+
+if sys.version_info[0] == 2:
+    from StringIO import StringIO  # pylint:disable=import-error
+else:
+    from io import StringIO
+
+
+def disable_repos(*args):
+    """Disable repos passed as ``args`` using ``subscription-manager repos
+    --disable``.
+
+    For example::
+
+        disable_repos('repo1', 'repo2')
+
+    Will run the command ``subscription-manager repos --disable repo1 repo2``.
+
+
+    """
+    run('subscription-manager repos --disable {0}'
+        .format(' '.join(['"{0}"'.format(repo) for repo in args])))
+
+
+def delete_custom_repos(**args):
+    """Delete repos files on ``/etc/yum.repos.d``.
+
+    All files that match ``<filename>.repo`` will be deleted. Be aware that
+    this task can delete repository files created by ``subscription-manager``
+    and other tools. But will raise ``ValueError`` if the repository name is
+    ``redhat``.
+
+    :raise: ``ValueError`` if repository name is 'redhat'.
+
+    """
+    for name in args:
+        name = name.rstrip('.repo')
+        if name == 'redhat':
+            raise ValueError('This task will not delete redhat.repo file.')
+        run('rm -f /etc/yum.repos.d/{0}.repo'.format(name), warn_only=True)
+
+
+def enable_repos(*args, **kwargs):
+    """Enable repos passed as ``args`` using ``subscription-manager repos
+    --enable``.
+
+    For example::
+
+        enable_repos('repo1', 'repo2')
+
+    Will run the command ``subscription-manager repos --enable repo1 repo2``.
+
+    """
+    run('subscription-manager repos --enable {0}'
+        .format(' '.join(['"{0}"'.format(repo) for repo in args])))
+
+
+def create_custom_repos(**kwargs):
+    """Create custom repofiles.
+
+    Each ``kwargs`` item will result in one repository file created. Where the
+    key is the repository filename and repository name, and the value is the
+    repository URL.
+
+    For example::
+
+        create_custom_repo(custom_repo='http://repourl.domain.com/path')
+
+    Will create a repository file named ``custom_repo.repo`` with the following
+    contents::
+
+        [custom_repo]
+        name=custom_repo
+        baseurl=http://repourl.domain.com/path
+        enabled=1
+        gpgcheck=0
+
+    """
+    for name, url in kwargs.items():
+        repo_file = StringIO()
+        repo_file.write(
+            '[{name}]\n'
+            'name={name}\n'
+            'baseurl={url}\n'
+            'enabled=1\n'
+            'gpgcheck=0\n'
+            .format(name=name, url=url)
+        )
+        put(local_path=repo_file,
+            remote_path='/etc/yum.repos.d/{0}.repo'.format(name))
+        repo_file.close()
+
+
+def enable_satellite_repos(cdn=False, disable_enabled=True):
+    """Enable repositories required to install Satellite 6
+
+    If ``disable_enabled`` is ``True``, then this task will first disable any
+    already enabled repository, including beaker ones.
+
+    """
+    if isinstance(cdn, str):
+        cdn = (cdn.lower() == 'true')
+    if isinstance(disable_enabled, str):
+        disable_enabled = (disable_enabled.lower() == 'true')
+
+    if disable_enabled is True:
+        disable_beaker_repos()
+        disable_repos('*')
+
+    repos = [
+        'rhel-{0}-server-rpms',
+        'rhel-server-rhscl-{0}-rpms',
+    ]
+
+    if cdn is True:
+        repos.append('rhel-{0}-server-satellite-6.0-rpms')
+
+    enable_repos(*[repo.format(distro_info()[1]) for repo in repos])
+
+    run('yum repolist')
+
+
+def disable_beaker_repos():
+    """Disable beaker repositories
+
+    If yum-config-manager is available this task will disable the repos, if not
+    it will move the beaker repo files to the running user home directory
+
+    """
+    # Clean up system if Beaker-based
+    result = run('which yum-config-manager', warn_only=True)
+    if result.succeeded:
+        run('yum-config-manager --disable "beaker*"')
+    else:
+        run('mv /etc/yum.repos.d/beaker-* ~/', warn_only=True)
+    run('rm -rf /var/cache/yum*')
+
+
+def manage_custom_repos(**kwargs):
+    """Enable or disable custom repositories.
+
+    The keyword key is the repository filename and the boolean value indicates
+    if it should enable if ``True`` or disable if ``False``.
+
+    """
+    for name, enable in kwargs.items():
+        repo_file = '/etc/yum.repos.d/{0}.repo'.format(name)
+        run('sed -i -e "s/^enabled=.*/enabled={0}/" {1}'.format(
+            1 if enable is True else 0,
+            repo_file
+        ))
--- a/automation_tools/utils.py
+++ b/automation_tools/utils.py
@@ -1,0 +1,81 @@
+"""Utilities tasks and functions"""
+from __future__ import print_function
+
+import re
+import sys
+
+from fabric.api import env, run
+
+
+def distro_info():
+    """Task which figures out the distro information based on the
+    /etc/redhat-release file
+
+    A ``(distro, major_version)`` tuple is returned if called as a function.
+    For RHEL X.Y.Z it will return ``('rhel', X)``. For Fedora X it will return
+    ``('fedora', X)``. Be aware that the major_version is an integer.
+
+    """
+    # Create/manage host cache
+    cache = env.get('distro_info_cache')
+    host = env['host']
+    if cache is None:
+        cache = env['distro_info_cache'] = {}
+
+    if host not in cache:
+        # Grab the information and store on cache
+        release_info = run('cat /etc/redhat-release', quiet=True)
+        if release_info.failed:
+            print('Failed to read /etc/redhat-release file')
+            sys.exit(1)
+
+        # Discover the distro
+        if release_info.startswith('Red Hat Enterprise Linux'):
+            distro = 'rhel'
+        elif release_info.startswith('Fedora'):
+            distro = 'fedora'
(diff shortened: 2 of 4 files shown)


automation_tools/repository.py
```diff
@@ -0,0 +1,126 @@
+"""Taks for managing repositories"""
+from __future__ import print_function
+
+import sys
+
+from automation_tools.utils import distro_info
+from fabric.api import put, run
+
+if sys.version_info[0] == 2:
+    from StringIO import StringIO  # (import-error) pylint:disable=F0401
```
Ichimonji10: You can write `pylint:disable=import-error`.

automation_tools/repository.py
```diff
@@ -0,0 +1,126 @@
...
+
+    For example::
+
+        disable_repos('repo1', 'repo2', repo3=False, repo4=True)
+
+    The above call will disable ``repo1`` and ``repo2`` using
+    ``subscription-manager repos --disable``. The ``repo3`` and ``repo4``
+    arguments will be considered custom repositories. For ``repo3`` this task
+    will disable it but will not delete de repository file, but for ``repo4``
+    it will delete de repository file.
```
Ichimonji10: `s/de/the/`

automation_tools/repository.py
```diff
@@ -0,0 +1,126 @@
...
+    If ``disable_enabled`` is ``True``, then this task will first disable any
+    already enabled repository, including beaker ones.
+
+    """
+    if isinstance(cdn, str):
+        cdn = (cdn.lower() == 'true')
+    if isinstance(disable_enabled, str):
+        disable_enabled = (disable_enabled.lower() == 'true')
+
+    if disable_enabled:
```
Ichimonji10: Using an explicit test like `if disable_enabled is True` may prevent some subtle bugs.

automation_tools/repository.py
elyezer: Thanks

automation_tools/repository.py
```diff
@@ -0,0 +1,126 @@
...
+    """Disable beaker repositories
+
+    If yum-config-manager is available this task will disable the repos, if not
+    it will move the beaker repo files to the running user home directory
+
+    """
+    # Clean up system if Beaker-based
+    result = run('which yum-config-manager', warn_only=True)
+    if result.succeeded:
+        run('yum-config-manager --disable "beaker*"')
```
Ichimonji10: Neat. This command succeeds even when there are no beaker repositories on the system.

```
# yum-config-manager | grep -i beaker
# ls /etc/yum.repos.d/beaker*
ls: cannot access /etc/yum.repos.d/beaker*: No such file or directory
# yum-config-manager --disable "beaker*"; echo $?
Loaded plugins: product-id
0
```

automation_tools/repository.py
elyezer: Thanks

automation_tools/utils.py
```diff
@@ -0,0 +1,81 @@
...
+import sys
+
+from fabric.api import env, run
+
+
+def distro_info():
+    """Task which figures out the distro information based on the
+    /etc/redhat-release file
+
+    A `(distro, major_version)` tuple is returned if called as a function. For
```
Ichimonji10: Double backticks denote inline code in a reStructuredText document. Single backticks denote inline code in a markdown document.

automation_tools/repository.py
```diff
@@ -0,0 +1,126 @@
...
+            run('sed -i -e "s/^enabled=.*/enabled=0/" {0}'.format(repo_file))
+
+
+def enable_repos(*args, **kwargs):
+    """Enable repos passed as ``args`` and create one repofile with each
+    ``kwargs``.
+
+    For example::
+
+        manage_repos('repo1', custom_repo='http://repourl.domain.com/path')
```
omaciel: s/manage_repos/enable_repos

Ichimonji10:
ACK pending comments.

automation_tools/repository.py
omaciel: :+1:

omaciel:
Talked with @elyezer and he agreed to break things down a bit more and the following methods will be created:

* ``enable_repos``
* ``disable_repos``
* ``create_custom_repo``
* ``delete_custom_repo``

This will make sure that the methods are a bit user friendlier and easy to understand.

automation_tools/repository.py
Ichimonji10: Good catch.

automation_tools/utils.py
elyezer: Thanks

automation_tools/repository.py
```diff
@@ -0,0 +1,156 @@
...
+def delete_custom_repos(**args):
+    """Delete repos files on ``/etc/yum.repos.d``.
+
+    All files that match ``<filename>.repo`` will be deleted, double check when
+    running this task because it can delete repository files created by
+    ``subscription-manager`` and other tools.
+
+    """
+    for name in args:
+        run('rm {0}'.format('/etc/yum.repos.d/{0}.repo'.format(name)),
```
omaciel: Maybe you should pass the ``-f`` flag here to make it non-interactive? Also, how about raising an exception if ``name == 'redhat'``? Finally, do we want to check ``name.endswith('.repo')``?

automation_tools/repository.py
```diff
@@ -0,0 +1,156 @@
...
+    # Clean up system if Beaker-based
+    result = run('which yum-config-manager', warn_only=True)
+    if result.succeeded:
+        run('yum-config-manager --disable "beaker*"')
+    else:
+        run('mv /etc/yum.repos.d/beaker-* ~/', warn_only=True)
+    run('rm -rf /var/cache/yum*')
+
+
+def manage_custom_repos(**kwargs):
```
Ichimonji10: Does it make sense to have this method (`manage_custom_repos`) in addition to the `enable_repos` and `disable_repos` methods? I took a look at the code and noticed that this method changes files in `/etc/yum.repod.d/` and the other two call `subscription-manager`. Is the difference here that `manage_custom_repos` works for any repo on the system and the other two only work on repos that are available through subscription-manager?

fabfile.py
```diff
@@ -33,10 +31,22 @@
...
+    disable_beaker_repos,
+    disable_repos,
+    enable_repos,
+    enable_satellite_repos,
+    manage_custom_repos,
+)  # flake8: noqa
+from automation_tools.utils import (
+    distro_info,
+    update_packages
+)  # flake8: noqa
```
Ichimonji10: `# flake8:noqa` causes flake8 to disable linting for the entire file, whereas `# noqa` causes flake8 to disable linting on a single line. I'd suggest dropping the `flake8: ` prefix. Beware that doing this may reveal linting issues that have been hidden.

automation_tools/utils.py
```diff
@@ -0,0 +1,81 @@
...
+    Possibly useful for when doing upgrades, etc.
+
+    """
+    if len(args) > 0:
+        arguments = ' '.join(args)
+    else:
+        arguments = ''
+
+    run(
+        'yum update -y {0}'.format(arguments),
```
Ichimonji10: Is it better to update all packages at once or to update packages one-by-one? I ask because, when `yum install -y …` returns 0 so long as at least one package successfully installs. Is there similar behaviour here, where `yum update -y …` returns 0 so long as at least one package successfully updates?

Ichimonji10:
ACK pending comments.

automation_tools/repository.py
omaciel: Your assumption is correct. Custom repositories cannot be enabled/disabled with ``subscription-manager``

automation_tools/repository.py
elyezer: Will update:

1. Pass the -f flag to rm
2. Does not allow to delete redhat repo
3. Will rstrip '.repo'

fabfile.py
elyezer: Cool, thanks.

automation_tools/utils.py
elyezer: That was a question I've done when this task was created, will do one at a time so the task fail and do any extra work letting the user decide what should be done.

fabfile.py
elyezer: Actually I will have to add `# noqa` for every line. I removed all noqa comments and flake8 shows the following:

```
./fabfile.py:2:1: F401 'add_repo' imported but unused
./fabfile.py:2:1: F401 'cdn_install' imported but unused
./fabfile.py:2:1: F401 'clean_rhsm' imported but unused
./fabfile.py:2:1: F401 'client_registration_test' imported but unused
./fabfile.py:2:1: F401 'create_personal_git_repo' imported but unused
./fabfile.py:2:1: F401 'downstream_install' imported but unused
./fabfile.py:2:1: F401 'errata_upgrade' imported but unused
./fabfile.py:2:1: F401 'fix_hostname' imported but unused
./fabfile.py:2:1: F401 'foreman_debug' imported but unused
./fabfile.py:2:1: F401 'install_errata' imported but unused
./fabfile.py:2:1: F401 'install_katello_agent' imported but unused
./fabfile.py:2:1: F401 'install_prerequisites' imported but unused
./fabfile.py:2:1: F401 'iso_download' imported but unused
./fabfile.py:2:1: F401 'iso_install' imported but unused
./fabfile.py:2:1: F401 'partition_disk' imported but unused
./fabfile.py:2:1: F401 'performance_tuning' imported but unused
./fabfile.py:2:1: F401 'product_install' imported but unused
./fabfile.py:2:1: F401 'remove_katello_agent' imported but unused
./fabfile.py:2:1: F401 'run_errata' imported but unused
./fabfile.py:2:1: F401 'setup_abrt' imported but unused
./fabfile.py:2:1: F401 'setup_ddns' imported but unused
./fabfile.py:2:1: F401 'setup_default_capsule' imported but unused
./fabfile.py:2:1: F401 'setup_default_docker' imported but unused
./fabfile.py:2:1: F401 'setup_email_notification' imported but unused
./fabfile.py:2:1: F401 'setup_fake_manifest_certificate' imported but unused
./fabfile.py:2:1: F401 'setup_firewall' imported but unused
./fabfile.py:2:1: F401 'setup_proxy' imported but unused
./fabfile.py:2:1: F401 'setup_vm_provisioning' imported but unused
./fabfile.py:2:1: F401 'subscribe' imported but unused
./fabfile.py:2:1: F401 'unsubscribe' imported but unused
./fabfile.py:2:1: F401 'update_basic_packages' imported but unused
./fabfile.py:2:1: F401 'upstream_install' imported but unused
./fabfile.py:2:1: F401 'vm_create' imported but unused
./fabfile.py:2:1: F401 'vm_destroy' imported but unused
./fabfile.py:2:1: F401 'vm_list' imported but unused
./fabfile.py:2:1: F401 'vm_list_base' imported but unused
./fabfile.py:40:1: F401 'create_custom_repos' imported but unused
./fabfile.py:40:1: F401 'delete_custom_repos' imported but unused
./fabfile.py:40:1: F401 'disable_beaker_repos' imported but unused
./fabfile.py:40:1: F401 'disable_repos' imported but unused
./fabfile.py:40:1: F401 'enable_repos' imported but unused
./fabfile.py:40:1: F401 'enable_satellite_repos' imported but unused
./fabfile.py:40:1: F401 'manage_custom_repos' imported but unused
./fabfile.py:49:1: F401 'distro_info' imported but unused
./fabfile.py:49:1: F401 'update_packages' imported but unused
```

Will continue skipping the entire file as we don't have any other code than F401.

Ichimonji10: :+1:

automation_tools/repository.py
Ichimonji10: Is there any reason that string interpolation is being performed twice here? I imagine it would be more straightforward to write this:

```python
run('rm /etc/yum.repos.d/{0}.repo'.format(name))
```

Ichimonji10:
ACK pending comment about double string interpolation.

automation_tools/repository.py
elyezer: Thank you! I really missed that and I don't know why I have written that way :open_mouth:

Ichimonji10:
:+1:

automation_tools/repository.py
```diff
@@ -0,0 +1,160 @@
...
+        cdn = (cdn.lower() == 'true')
+    if isinstance(disable_enabled, str):
+        disable_enabled = (disable_enabled.lower() == 'true')
+
+    if disable_enabled is True:
+        disable_beaker_repos()
+        disable_repos('*')
+
+    # Disable yum plugin for sub-man
+    run('sed -i -e "s/^enabled.*/enabled=0/" '
```
omaciel: Since this task may be called several times, does it make sense to always run this? Will it 'fail' if called many times?

elyezer: We can improve this by searching for enabled equals 1. And the file will be edited only if enabled = 1

elyezer: Seems that this is not more required, will drop this code.

omaciel:
APT

Issue #174, opened 2015-04-22
TypeError right after running satellite5_iso_install task


lpramuk:
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: Installation complete.
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: Visit https://${FQDN} to create the Red Hat Satellite administrator account.
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: 

[dell-pe2950-01.lab.eng.rdu.redhat.com] run: yum -y update
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: Loaded plugins: product-id, rhnplugin, security
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: This system is receiving updates from RHN Classic or RHN Satellite.
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: Setting up Update Process

[dell-pe2950-01.lab.eng.rdu.redhat.com] out: No Packages marked for Update
[dell-pe2950-01.lab.eng.rdu.redhat.com] out: 

Traceback (most recent call last):
  File "/home/jenkins/shiningpanda/jobs/35403905/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/main.py", line 743, in main
    *args, **kwargs
  File "/home/jenkins/shiningpanda/jobs/35403905/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 384, in execute
    multiprocessing
  File "/home/jenkins/shiningpanda/jobs/35403905/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 274, in _execute
    return task.run(*args, **kwargs)
  File "/home/jenkins/shiningpanda/jobs/35403905/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 174, in run
    return self.wrapped(*args, **kwargs)
  File "/home/jenkins/workspace/satellite5-installer/automation_tools/__init__.py", line 1011, in product_install
    )[host])
TypeError: 'NoneType' object is not iterable

lpramuk:
Solved as PR #175 was merged

PR #175, opened 2015-04-23
fix sat5_iso_install task

lpramuk:
fixes Issue #174 TypeError right after running satellite5_iso_install task

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -893,6 +893,8 @@ def satellite5_iso_install(
             '--re-register --run-updater=yes --enable-tftp=yes')
         run('yum -y update')
 
+    return {}
+
 
 def sam_upstream_install(admin_password=None):
     """Task to install SAM nightly using katello-deploy script"""


elyezer:
ACK, this will fix the issue, but seems that we will need further refactor to make the product_install task install Satellite 5. Or maybe will be better to split to a new task.

automation_tools/__init__.py
```diff
@@ -888,6 +888,9 @@ def satellite5_iso_install(
             '--re-register --run-updater=yes --enable-tftp=yes')
         run('yum -y update')
 
+    installer_options = {}
+    return installer_options
```
Ichimonji10: How about this:

```python
return {}
```

Ichimonji10:
#174 does not clearly describe what the source of the traceback is. Without more information about the source of the problem, my gut feeling tells me that this simply papers over a deeper problem. @lpramuk, can you spend some time digging into this issue and detail your findings? If nothing else, I'd appreciate knowing what's going on.

lpramuk:
Like elyezer said before we will need further refactor to make the product_install.
Currently we used iso_install task as a teplate for sat5_iso_install task, so it has to mimics what it returns to product_install task.
In the future we may consider to create sole sat5_product_install not to obey sat6 existing constructs....

Currently sat5_iso_install produces traceback, so let's consider this as a quick fix before further refactoring...

Ichimonji10:
> Currently sat5_iso_install produces traceback, so let's consider this as a quick fix before any further refactoring...

I'm wary of papering over problems with quick fixes. I understand that this change would fix the traceback, and have asked for more information with that in mind.

> Currently we used iso_install task as a teplate for sat5_iso_install task, so it has to mimics what it returns to product_install task.

Aha. Now I understand.

ACK pending comment.

omaciel:
@lpramuk I agree with what @Ichimonji10  said about 'quick fixes', so can you file a new Github issue with information about what the real fix should be and assign it to yourself?

lpramuk:
Guys I'm on your side too. I didn't cause the need for the quick fix.
I delivered what was omitted by someone else...

Lets agree on creating new issue for code refactoring of product_install to move out of this...

omaciel:
@lpramuk we **are** on the same side and my comment was the agreement that we want a separate issue to track this :)

lpramuk:
we want it we get it - new Issue #177

elyezer:
@lpramuk ping me once you incorporate the @Ichimonji10's comment se we can merge this.

elyezer:
APT

Ichimonji10:
ACK

PR #178, opened 2015-04-27
Add setup_scap_client task

elyezer:
This task will install the puppet-foreman_scap_client package. Also make
product_install run setup_scap_client task for every installation,
Robottelo is about to have tests that will rely on the
puppet-foreman_scap_client package being installed.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -418,6 +418,11 @@ def setup_abrt():
     run('abrt-auto-reporting enabled')
 
 
+def setup_scap_client():
+    """Task to setup puppet-foreman_scap_client."""
+    run('yum -y install puppet-foreman_scap_client', warn_only=True)
+
+
 def vm_create():
     """Task to create a VM using snap-guest based on a ``SOURCE_IMAGE`` base
     image.
@@ -1006,6 +1011,8 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         sam=distribution.startswith('sam'),
         **installer_options
     )
+
+    execute(setup_scap_client, host=host)
 
     certificate_url = certificate_url or os.environ.get(
         'FAKE_MANIFEST_CERT_URL')


automation_tools/__init__.py
```diff
@@ -418,6 +418,11 @@ def setup_abrt():
     run('abrt-auto-reporting enabled')
 
 
+def setup_scap_client():
```
elyezer: A task is required in order to use the `execute` on `product_install`. `execute` allows specify the host where the commands should be run and we need this to run the command on a machine or on a VM (if creating one).

automation_tools/__init__.py
```diff
@@ -1007,6 +1012,8 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         **installer_options
     )
 
+    execute(setup_scap_client, host=host)
```
elyezer: Every time execute is called on `product_install` it should specify `host=host`. It could be a machine when called with `-H root@server.com` or can be a VM when creating one and the `-H root@provisioning-host.com` will be the provisioning machine, after creating the VM the host will be updated to run on the VM instead of the provisioning host.

automation_tools/__init__.py
```diff
@@ -418,6 +418,11 @@ def setup_abrt():
     run('abrt-auto-reporting enabled')
 
 
+def setup_scap_client():
+    """Task to setup puppet-foreman_scap_client."""
+    run('yum -y puppet-foreman_scap_client', warn_only=True)
```
jyejare: I think before directly running installation. we can make sure that scap_client is available for install from yum. By doing yum list.

jyejare:
ACK pending Comments

automation_tools/__init__.py
elyezer: We needed to do that for ABRT because the package was not available on RHEL 6 repos. This is not needed and the warn_only will not let the installation fail.

jyejare:
ACK. Ignore my comment.

automation_tools/__init__.py
kbidarkar: shouldn't it be "yum -y install puppet-foreman_scap_client" , I think we are missing "install", is it so ?

kbidarkar:
ACK

Issue #177, opened 2015-04-27
refactor product_install task for Sat5 


lpramuk:
We need further refactor to make the product_install task install Satellite 5
Currently we used iso_install task as a teplate for sat5_iso_install task, so it has to mimics what it returns to product_install task.

elyezer:
The design decision behind product_install and all *_install tasks was that every install task should return a map for installer options, then the katello_installer task grab that information and run the install command. One way to refactor is to adopt the same approach for Satellite 5 related tasks or think a different approach and make product_install handle it if doing a Satellite 5 installation.

Ichimonji10:
Thanks for the explanation, @elyezer.

lpramuk:
I'd go for separate satellite5_product_install to avoid frequent if conditions for sat5 vs. sat6

elyezer:
:+1:

lpramuk:
Submitted PR #180 
Oops, looks like I missed elyezer last comment, yet I'd like to incorporate it in.

lpramuk:
Resolved as PR #180 was merged.

PR #180, opened 2015-05-06
satellite5 installation task refactoring

lpramuk:
created separate satellite5_product_install task that utilizes satellite5_installer (that is renamed and modified satellite5_iso_install task)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -821,84 +821,6 @@ def iso_install(
         return installer_options
 
 
-def satellite5_iso_install(
-        admin_password=None, check_sigs=False, run_katello_installer=True):
-    """Installs Satellite 5 from an ISO image.
-
-    The following environment variables affect this command:
-
-    """
-    if isinstance(check_sigs, str):
-        check_sigs = (check_sigs.lower() == 'true')
-
-    if admin_password is None:
-        admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
-
-    iso_url = os.environ.get('ISO_URL') or os.environ.get('BASE_URL')
-    if iso_url is None:
-        print('Please provide a valid URL for the ISO image.')
-        sys.exit(1)
-
-    # Download the ISO
-    print('Downloading ISO...')
-    iso_download(iso_url)
-    # Create a 'check-out' folder, mount ISO to it...
-    run('mkdir -p ISO')
-    run('mount *.iso ISO -t iso9660 -o loop')
-
-    # prepare the answer file
-    opts = {
-        'admin-email': os.environ.get('ADMIN_EMAIL', ''),
-        'rhn-username': os.environ.get('RHN_USERNAME', ''),
-        'rhn-password': os.environ.get('RHN_PASSWORD', ''),
-        'rhn-profile-name': os.environ.get('RHN_PROFILE', ''),
-        'rhn-http-proxy': os.environ.get('RHN_HTTP_PROXY', ''),
-        'rhn-http-proxy-username':
-        os.environ.get('RHN_HTTP_PROXY_USERNAME', ''),
-        'rhn-http-proxy-password':
-        os.environ.get('RHN_HTTP_PROXY_PASSWORD', ''),
-        'ssl-set-org': os.environ.get('SSL_SET_ORG', ''),
-        'ssl-set-org-unit': os.environ.get('SSL_SET_ORG_UNIT', ''),
-        'ssl-set-city': os.environ.get('SSL_SET_CITY', ''),
-        'ssl-set-state': os.environ.get('SSL_SET_STATE', ''),
-        'ssl-set-country': os.environ.get('SSL_SET_COUNTRY', ''),
-        'ssl-password': os.environ.get('SSL_PASSWORD', ''),
-        'satellite-cert-file': os.environ.get('SATELLITE_CERT_FILE', '')
-    }
-    run(
-        'cat <<EOF > /answers.txt\n'
-        'admin-email={admin-email}\n'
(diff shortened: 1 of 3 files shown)


automation_tools/__init__.py
```diff
@@ -1026,6 +947,129 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+def satellite5_product_install(create_vm=False, selinux_mode=None):
+    """Task which installs only satellite5 product.
+
+    If ``create_vm`` is True then ``vm_destroy`` and ``vm_create`` tasks will
+    be run. Make sure to set the required environment variables for those
+    tasks. Also, if one of the ``setup_ddns`` required environment variables
+    is set then that task will run.
+
+    :param bool create_vm: creates a virtual machine and then install the
+        product on it. Default: False.
```
Ichimonji10: Documentation generators such as Sphinx typically are smart enough to examine the arguments to a method and show the default in the generated docstring. For example, see [this method](http://nailgun.readthedocs.org/en/latest/api/nailgun.html#nailgun.config.BaseServerConfig.get).

automation_tools/__init__.py
```diff
@@ -1026,6 +947,129 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+        'satellite-cert-file=/root/SATCERT\n'
+        'enable-tftp=yes\n'
+        'EOF\n'.format(**opts)
+    )
+
+    # download a certificate
+    print('Downloading Certificate...')
+    run('wget -nv -O /root/SATCERT {satellite-cert-url}'.format(**opts))
+    # ...and run the installer script.
+    with cd('/root/ISO'):
```
Ichimonji10: Earlier, the directory `ISO` was created. Here, `/root/ISO` is being used. I would suggest just using `ISO`, especially in case a non-root user executes these instructions. The use of the root user should not be assumed if possible, IMO.

automation_tools/__init__.py
```diff
@@ -1026,6 +947,129 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+    iso_url = os.environ.get('ISO_URL')
+    if iso_url is None:
+        print('Please provide a valid URL for the ISO image.')
+        sys.exit(1)
+
+    # Download and mount the ISO
+    print('Downloading ISO...')
+    iso_download(iso_url)
+    run('mkdir -p ISO')
+    run('mount -t iso9660 -o loop *.iso ISO')
```
Ichimonji10: Do you ever unmount and/or remove the ISO directory?

Ichimonji10:
Is this task going to be moved in to a separate directory as per [this comment](https://github.com/SatelliteQE/automation-tools/issues/177#issuecomment-98822753)? If so, then please mark this PR as `[DO NOT MERGE]` until that is done.

automation_tools/__init__.py

automation_tools/__init__.py

automation_tools/__init__.py
elyezer: In the Satellite 6 ISO job we don't unmount the directory, I don't know the reason but maybe it will be used later. Anyway if we discover that the ISO can be unmounted then we should do the same for Satellite 6 task.

fabfile.py
```diff
@@ -19,6 +19,7 @@
     product_install,
     remove_katello_agent,
     run_errata,
+    satellite5_product_install,
```
elyezer: I think we can list `satellite5_installer` here too.

elyezer:
ACK, pending comments.

fabfile.py
lpramuk: Ok, then I see more tasks missing, like katello-installer task too.

automation_tools/__init__.py
Ichimonji10: OK. Sounds like something for a future investigation.

fabfile.py
elyezer: Good catch, but other missing tasks should be added in a separated commit or even another PR.

lpramuk: sure i'm not gonna include these changes in PR

lpramuk:
- comments incorporated into PR
- code splitted into separate module

automation_tools/satellite5.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    iso_url = os.environ.get('ISO_URL')
+    if iso_url is None:
+        print('Please provide a valid URL for the ISO image.')
+        sys.exit(1)
+
+    # Download and mount the ISO
+    print('Downloading ISO...')
+    iso_download(iso_url)
+    run('umount ISO', warn_only=True)
```
Ichimonji10: :+1:

Ichimonji10:
ACK

automation_tools/satellite5.py
```diff
@@ -0,0 +1,141 @@
...
+
+Many commands are affected by environment variables. Unless stated otherwise,
+all environment variables are required.
+
+"""
+from __future__ import print_function
+import os
+import sys
+
+from automation_tools import install_prerequisites, iso_download, setenforce, \
```
elyezer: Please avoid using `\`, the below example is how we would format that.

```python
from automation_tools import (
    install_prerequisites,
    iso_download,
    setenforce,
    setup_ddns,
    subscribe,
    vm_create,
    vm_destroy,
)
```

elyezer:
ACK, pending last comment format.

automation_tools/satellite5.py
lpramuk: fixed.

elyezer:
APT

elyezer:
Karma given.

PR #182, opened 2015-05-07
Specify the tftp-server-name as the sat6 hostname

kbidarkar:
*) currently the tftp-server-name is picked up as nil or the
   docker ip, as default if not specified.
*) so, it's safe to explicitly specify
   --tftp-server-name <hostname>


sghai:
ACK..good catch

elyezer:
ACK, thank you.

PR #183, opened 2015-05-12
Add workaround for Java on RHEL 6.6

elyezer:
This workaround is needed in order to install downstream on RHEL 6.6
without getting an error. If Java is not installed will be selected the
1.8.0 one which make the build fail, but the 1.7.0 will make it work
properly.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -721,6 +721,13 @@ def downstream_install(admin_password=None, run_katello_installer=True):
         remote_path='/etc/yum.repos.d/satellite.repo')
     satellite_repo.close()
 
+    info = distro_info()
+    if info == ('rhel', 6, 6):
+        # For default java-1.8.0-openjdk will be installed on RHEL 6.6 but it
+        # makes the katello-installer fail. Install java-1.7.0-openjdk which is
+        # the recommended version for RHEL 6.6.
+        run('yum install -y java-1.7.0-openjdk')
+
     # Install required packages for the installation
     run('yum install -y katello')
 


automation_tools/__init__.py
```diff
@@ -721,6 +721,13 @@ def downstream_install(admin_password=None, run_katello_installer=True):
         remote_path='/etc/yum.repos.d/satellite.repo')
     satellite_repo.close()
 
+    info = distro_info()
+    if info == ('rhel', 6, 6):
+        # For default java-1.8.0-openjdk will be installed on RHEL 6.6 but it
+        # makes the katello-installer fail. Install java-1.7.0-openjdk which is
+        # the recommended version for RHEL 6.6.
+        run('yum install -y java-1.7.0-openjdk')
+
```
sthirugn: I did the following workaround to get installation in rhel 6.6 going, any reason you are using only one of the three steps which I mentioned below?

```sh
yum -y install java-1.7.0-openjdk
alternatives --config java 
--> (selected 1.7.0 here)
# rpm -e java-1.8.0-openjdk-headless-1.8.0.45-28.b13.el6_6.x86_64 java-1.8.0-openjdk-1.8.0.45-28.b13.el6_6.x86_64
```

elyezer: If java 1.7.0 is already installed there is no need to deal with the 1.8.0 because the installer will not install it.

The steps I do when fixing it are:

```
yum remove java-1.8.0-openjdk
yum install java-1.7.0-openjdk
yum install katello
katello-installer ...
```

sthirugn: Thank you.  What about the following:
```sh
# alternatives --config java 
```

elyezer: I don't have to run that in order to make the installer work.

sthirugn: Interesting, lets get this merged and talk about `alternatives` later.

sthirugn:
ACK

sthirugn:
Merging with one ACK to get automation going.

PR #190, opened 2015-06-09
Open port 8000 for Capsule communication

elyezer:
Now Capsule communication requires that port 8000 is open in order to
work. Even for the default capsule.


omaciel:
ACK

JacobCallahan:
APT

PR #197, opened 2015-07-01
Update requirements for python-novaclient

jyejare:
Just an update in requirements.txt for python-novaclient.Which triggered travis build failed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 Fabric
-novaclient
+python-novaclient


Ichimonji10:
This is an improvement, but not a fix for build failures like [this](https://travis-ci.org/SatelliteQE/robottelo). ACK, but beware that this only fixes a bug in automation-tools itself and a follow-up PR will need to be submitted.

Ichimonji10:
Also see #198.

PR #198, opened 2015-07-01
Fix dependency specification

Ichimonji10:
* Make `requirements.txt` reference `setup.py`. This means that there is a
  single authoritative list of dependencies.
* Make Travis install dependencies as part of the test process.
* Move all requirements to `setup.py`. This should let pip install
  automation-tools' dependencies when automation-tools is itself a dependency.

Here's a log from my own system, indicating that the fix works correctly:

    $ virtualenv --python python2 env
    $ source env/bin/activate
    $ git clone git@github.com:SatelliteQE/robottelo.git
    $ cd robottelo
    $ vim requirements.txt  # -e /home/ichimonji10/code/automation-tools
    $ pip install -r requirements.txt
    $ python  # import novaclient

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 install:
     - pip install flake8
+    - pip install -r requirements.txt
 script:
     - flake8 .
 notifications:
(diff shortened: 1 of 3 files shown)


rplevka:
ACK :+1:

sthirugn:
ACK

elyezer:
:+1:

PR #208, opened 2015-07-23
Fixes the update packages bug

sthirugn:
- The update_packages function tries to update the hypervisor rather than the intended vm due to fab context change

Note: I am not able to reproduce the bug locally and this happens only in the automation host-vm setup.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -918,7 +918,9 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         beta=distribution.endswith('beta'),
         host=host
     )
-    execute(update_packages, warn_only=True)
+
+    # Update the machine
+    update_packages(warn_only=True)
 
     if distribution in ('satellite6-downstream', 'satellite6-iso'):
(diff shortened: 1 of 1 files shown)


Ichimonji10:
ACK

JacobCallahan:
ACK

automation_tools/__init__.py
```diff
@@ -918,7 +918,9 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         beta=distribution.endswith('beta'),
         host=host
     )
-    execute(update_packages, warn_only=True)
+
+    # Update the machine
+    update_packages(warn_only=True)
```
elyezer: The right fix would be something like, `execute(update_packages, warn_only=True, host=host)`.

sthirugn: Yeah my fix was incorrect and the automation failed agan.  I investigarted run and it does not take host parameter.  So your suggested fix will fail as well.
I am testing this instead
```py
with settings(warn_only=True, host_string=host):
        update_packages()
```

PR #209, opened 2015-07-23
Really fix bug found on #208

elyezer:
Call `update_packages` using `execute` and pass the `host` in order to
run on the provisioned virtual machine, if provisioning one, or, on the
defined machine, when not.


omaciel:
ACK

Ichimonji10:
ACK

sthirugn:
@elyezer thank you :) I retriggered and it works fine now.

PR #210, opened 2015-07-31
Update setup_default_docker task

elyezer:
Match the steps recommended on the Docker prerequisites section from the
docs.

Drop pulling busybox image and running the docker ps command if any
issue happens the manage_daemon will fail.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -192,10 +192,8 @@ def setup_default_docker():
     else:
         run('yum install -y docker-io', warn_only=True)
 
-    # Docker should run as the ``apache`` user
-    for group in ('docker', 'dockerroot'):
-        if run('id -g {0}'.format(group), quiet=True).succeeded:
-            run('usermod -aG {0} apache'.format(group))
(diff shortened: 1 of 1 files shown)


omaciel:
**NOTE**: this will now automatically configure Docker to be used either as internal (unix socket) and external (http) compute resource.

ACK

automation_tools/__init__.py
```diff
@@ -218,13 +217,6 @@ def setup_default_docker():
     # https://github.com/docker/docker/issues/2758
     manage_daemon('restart', 'docker', pty=(os_version >= 7))
 
-    # Check that things look good
-    run('docker ps')
```
sthirugn: We can leave this just to make sure that docker is really installed? (may be with warn_only=True)

sthirugn:
ACK pending comment

automation_tools/__init__.py
elyezer: @sthirugn as the commit message states, manage_daemon will fail if docker is not installed.

sthirugn: ok :+1:

PR #211, opened 2015-08-04
Add workaround for qdrouterd to listen to IPv6

abalakh:
By default, `qdrouterd` is only listening to IPv4.
Content-hosts, created from VM-images in our automation are trying to connect through IPv6.
There's already a bug opened for `qdrouterd` to make it listen to IPv6 - [BZ1219902](https://bugzilla.redhat.com/show_bug.cgi?id=1219902).
As there're some content-host tests upcoming, i suggest applying this workaround, which basically editі the `qdrouterd` config and restarts the service at the end of sat installation.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -964,6 +964,19 @@ def product_install(distribution, create_vm=False, certificate_url=None,
             host=host
         )
 
+    execute(fix_qdrouterd_listen_to_ipv6, host=host)
+
+
+def fix_qdrouterd_listen_to_ipv6():
+    """Configure qdrouterd to listen to IPv6 instead of IPv4.
+
+    Workaround for BZ #1219902.
+
+    """
+    run('sed -i -e "0,/addr: 0.0.0.0/s/addr: 0.0.0.0/addr: ::/" '
+        '/etc/qpid-dispatch/qdrouterd.conf')
+    manage_daemon('restart', 'qdrouterd')
+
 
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -964,6 +964,19 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+    execute(fix_qdrouterd_listen_to_ipv6, host=host)
+
+
+def fix_qdrouterd_listen_to_ipv6():
+    """Configure qdrouterd to listen to IPv6 instead of IPv4.
+
+    Workaround for BZ #1219902.
+
+    """
+    run('sed -i -e "s|addr: 0.0.0.0|addr: ::|" '
```
sthirugn: There are two instances of `addr: 0.0.0.0` in this config - one corresponds to port `5647` and the other corresonding to `5646`.  When testing, I fixed the port `5647` only.  Having said that, I am not sure if fixing both may create problem or not.  May be we can merge this and address any issues, if they occur in future.

sthirugn:
ACK - but please review my non-blocker comment.

automation_tools/__init__.py
abalakh: I tried replacing both and it was working. But, of course, I'm not sure of any new issues either. I'll better replace only the first occurrence, thanks for suggestion.

elyezer:
ACK

PR #212, opened 2015-08-04
Update setup_vm_provisioning task

### LINKED ISSUE #169: setup_vm_provisioning checks for a nested virtualization support prior to virtualization support itself
`setup_vm_provisioning` first checks for a nested virtualization support on the machine by checking the output of _/sys/module/kvm_intel/parameters/nested_ file.
If not supported, it tries to enable it and prompts user to reboot the host.
After that, it is supposed to proceed in code and start checking for virtualization support. This stage cannot be reached in case no virtualization support exists on the host thus I believe these checks should be swapped.

elyezer:
Check for virtualization support before checking for nested
virtualization.

Closes #169

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -538,6 +538,12 @@ def setup_vm_provisioning(interface=None):
         print('A network interface is required')
         sys.exit(1)
 
+    # Check for virtualization support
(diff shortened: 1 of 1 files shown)


sthirugn:
ACK

jyejare:
ACK

PR #214, opened 2015-08-06
Upgrade Process Updates

jyejare:
Some Updates regarding overall upgrade process and on few updates related to openstack instance specific.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1640,6 +1640,7 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
         openstack_client.floating_ip_pools.list()[0].name
     )
     # Create instance from the given parameters
+    print('Creating new Openstack instance {0}'.format(instance_name))
     instance = openstack_client.servers.create(
         name=instance_name,
         image=image.id,
@@ -1686,6 +1687,9 @@ def delete_openstack_instance(instance_name):
         ))
         return
     instance.delete()
+    print('The instance {0} has been deleted from Openstack.'.format(
+        instance_name
+    ))
 
 
 def satellite6_upgrade(admin_password=None):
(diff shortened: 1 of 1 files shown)


sghai:
ACK.. changes looks good. thanks

automation_tools/__init__.py
```diff
@@ -1726,6 +1729,8 @@ def satellite6_upgrade(admin_password=None):
     # Stop katello services, except mongod
     run('katello-service stop')
     run('service-wait mongod start')
+    if major_ver == 7:
+        run('service tomcat stop')
```
Ichimonji10: If this bit of code only runs on RHEL 7, then use systemd.

    run('systemctl stop tomcat')

Ichimonji10:
ACK pending comment.

automation_tools/__init__.py
elyezer: Use manage_daemon helper. It will ensure to run the right service management utility depending on the rhel version.

automation_tools/__init__.py
```diff
@@ -1640,6 +1640,7 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
         openstack_client.floating_ip_pools.list()[0].name
     )
     # Create instance from the given parameters
+    print("Creating new Openstack instance {0}".format(instance_name))
```
elyezer: nitpick s/"/'

automation_tools/__init__.py
```diff
@@ -1686,6 +1687,9 @@ def delete_openstack_instance(instance_name):
         ))
         return
     instance.delete()
+    print("The instance {0} has been deleted from Openstack.".format(
```
elyezer: nitpick s/"/'

automation_tools/__init__.py
Ichimonji10: :+1:

jyejare: This is the standard procedure/command we are following in Upgrade procees, which is also documented. So don't want to make any changes in this. So let it be.

automation_tools/__init__.py
jyejare: Done.

automation_tools/__init__.py
jyejare: Done.

jyejare:
Updated code as per comments. Please merge/re-review accordingly.

automation_tools/__init__.py
Ichimonji10: Ahh, I see. Well, as long as it works.

Ichimonji10:
Looks OK to me. I'll let someone else verify, though.

sghai:
Merging this as @jyejare needs the changes in for another related task.

PR #216, opened 2015-08-10
Fix for fabric to prevent losing of host context

sthirugn:
Fabric loses the host context during katello-service call.  Since this can't be tested locally, I am not sure yet if this fix will really fix the problem.  I am going to attempt to merge this to get the automation going.


sthirugn:
Merged to retrigger automation

PR #225, opened 2015-08-20
Static delay removed to resolve the DNS

jyejare:
In this PR, the static delay has been removed/replaced by host_pings method

Earlier there was fix wait time of 10 minutes to resolve the DNS, which is now replaced by pinging the IP and then pinging the hostname retrieved by IP.

This surely makes run upgrade automation faster. :)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -11,6 +11,7 @@ import socket
 import sys
 import time
 import novaclient
+import subprocess
 from re import search
 from urlparse import urlsplit
 
@@ -1622,6 +1623,29 @@ def setenforce(mode):
     run('setenforce {0}'.format(mode))
 
 
+def host_pings(host, attempts=200):
+    """This ensures the given IP/hostname pings succesfully.
+
+    :param host: A string. The IP or hostname of host.
+    :param attempts: An integer.
+        The maximum number of attempts to ping the host.
+
+    """
+    for attempt in range(attempts):
+        command = subprocess.Popen(
+            'ping -c1 {0}; echo $?'.format(host),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=True
+        )
+        output = command.communicate()[0]
+        # Checking the return code of ping is 0
+        if int(output.split()[-1]) == 0:
+            break
+        else:
+            time.sleep(5)
+
+
 def get_hostname_from_ip(ip):
     """Retrives the hostname by logging into remote machine by IP.
     Specially for the systems who doesnt support reverse DNS.
@@ -1673,7 +1697,7 @@ def get_openstack_client():
 
 def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
     """Creates openstack Instance from Image and Assigns a floating IP
-    to instance.
+    to instance. Also It ensures that instance is ready for testing.
 
     :param instance_name: A string. Openstack Instance name to create.
     :param image_name: A string. Openstack image name from which instance
@@ -1725,9 +1749,16 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
         except novaclient.exceptions.BadRequest:
             time.sleep(5)
     # Wait till DNS resolves the IP
-    time.sleep(600)
-    # Getting Hostname from IP
-    env['instance_host'] = get_hostname_from_ip(str(floating_ip.ip))
+    print('Pinging the Host by IP:{0} ..........'.format(str(floating_ip.ip)))
+    host_pings(str(floating_ip.ip))
+    print('SUCCESS !! The given IP has been pinged!!\n')
+    print('Now, Getting the hostname from IP......\n')
+    hostname = get_hostname_from_ip(str(floating_ip.ip))
+    env['instance_host'] = hostname
+    print('Pinging the Hostname:{0} ..........'.format(hostname))
+    host_pings(hostname)
+    print('SUCCESS !! The obtained hostname from IP is pinged !!')
+    print('The Instance is ready for further Testing .....!!')
 
 
 def delete_openstack_instance(instance_name):
@@ -1779,6 +1810,7 @@ def satellite6_upgrade(admin_password=None):
     # Removing rhel-released and rhel-optional repo
     run('rm -rf /etc/yum.repos.d/rhel-{optional,released}.repo')
     # Update the packages
+    print('Wait till Packages update ... ')
     update_packages(quiet=True)
     # Setting Satellite61 Repos
     major_ver = distro_info()[1]
@@ -1803,6 +1835,7 @@ def satellite6_upgrade(admin_password=None):
     # yum cleaning all
     run('yum clean all', warn_only=True)
     # Updating the packages again after setting sat6 repo
+    print('Wait till packages update ... ')
     update_packages(quiet=True)
     # Upgrading Katello installer
     run('katello-installer --upgrade')


automation_tools/__init__.py
```diff
@@ -1622,6 +1623,30 @@ def setenforce(mode):
...
+        sys.exit(1)
+    while True:
+        command = subprocess.Popen(
+            'ping -c1 {0}'.format(host),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            shell=True
+        )
+        output = command.communicate()[0]
+        if '1 received' in output:
```
Ichimonji10: How about using return codes instead?

```
       If ping does not receive any reply packets at all it will exit with  code  1.  If  a
       packet  count  [ping(8)]  and  deadline [ping(8)] are both specified, and fewer than
       count [ping(8)] packets are received by the time the deadline [ping(8)] has arrived,
       it  will  also  exit with code 1.  On other error it exits with code 2. Otherwise it
       exits with code 0. This makes it possible to use the exit code to see if a  host  is
       alive or not.
```

automation_tools/__init__.py
```diff
@@ -1622,6 +1623,30 @@ def setenforce(mode):
...
+def host_pings(host):
+    """This ensures the given IP/hostname pings succesfully.
+
+    :param host: A string. The IP or hostname of host.
+
+    """
+    if host is None:
+        print("The hostname/IP is not provided to ping. \
+        Terminating the execution !!")
+        sys.exit(1)
```
Ichimonji10: Drop this `if host is None:` block and just make sure to call the method correctly.

automation_tools/__init__.py
```diff
@@ -1673,7 +1698,7 @@ def get_openstack_client():
 
 def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
     """Creates openstack Instance from Image and Assigns a floating IP
-    to instance.
+    to instance. Also It ensures that istance is ready for testing.
```
Ichimonji10: `s/istance/instance/`?

automation_tools/__init__.py
```diff
@@ -1622,6 +1623,30 @@ def setenforce(mode):
     run('setenforce {0}'.format(mode))
 
 
+def host_pings(host):
```
Ichimonji10: This function runs indefinitely if the pinged host never responds. How about a time-out? Please read `ping`'s man page and see if it has the requisite functionality built-in.

```
       Stop after sending count ECHO_REQUEST
              packets.  With  deadline  [ping(8)]  option,  ping waits for count ECHO_REPLY
              packets, until the timeout expires.
```

automation_tools/__init__.py
```diff
@@ -1725,9 +1750,16 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
         except novaclient.exceptions.BadRequest:
             time.sleep(5)
     # Wait till DNS resolves the IP
-    time.sleep(600)
-    # Getting Hostname from IP
-    env['instance_host'] = get_hostname_from_ip(str(floating_ip.ip))
+    print("Pinging the Host by IP..........")
```
Ichimonji10: Please note which IP address is being pinged.

automation_tools/__init__.py
```diff
@@ -1725,9 +1750,16 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
...
+    host_pings(str(floating_ip.ip))
+    print("SUCCESS !! The given IP has been pinged!!\n")
+    print("Now, Getting the hostname from IP......\n")
+    hostname = get_hostname_from_ip(str(floating_ip.ip))
+    env['instance_host'] = hostname
+    print("Pinging the Hostname..........")
```
Ichimonji10: Please note which hostname is being pinged.

automation_tools/__init__.py
```diff
@@ -1803,6 +1836,7 @@ def satellite6_upgrade(admin_password=None):
     # yum cleaning all
     run('yum clean all', warn_only=True)
     # Updating the packages again after setting sat6 repo
+    print("Wait till packages update ... ")
```
Ichimonji10: Please use single quotes throughout this pull request.

automation_tools/__init__.py

automation_tools/__init__.py
```diff
@@ -1622,6 +1623,30 @@ def setenforce(mode):
...
+    """This ensures the given IP/hostname pings succesfully.
+
+    :param host: A string. The IP or hostname of host.
+
+    """
+    if host is None:
+        print("The hostname/IP is not provided to ping. \
+        Terminating the execution !!")
+        sys.exit(1)
+    while True:
```
elyezer: This could cause a deadlock if the server never goes up. What about adding a retry number?

Also do we have the mean time for a DNS publishing?

automation_tools/__init__.py

elyezer:
ACK, pending comments.

automation_tools/__init__.py
jyejare: @elyezer , Mean time for RHEL 7 is around 7-8 minutes and for RHEL 6 it is around 12-14 minutes.

automation_tools/__init__.py
jyejare: So in this case if the exit status code anything other than status code 2 means the host is alive.

Where,
0, means the numbers of packets transmitted == no. of packets received, This means the host is alive
1, means the numbers of packets transmitted >= no. of packets received, This also means the host is alive but there is packet loss just because the deadline has reached.
2, means 100% packet loss , the host is not alive at all.

So you think this approach is correct that I should keep on pining the host until the status code is 0/1.

automation_tools/__init__.py
jyejare: I am seriously not understanding how do I use timeout here. Because I have to loop the ping until the status code is 0/1.

automation_tools/__init__.py
Ichimonji10: You're currently looping and using `-c 1`, so you can simply check for a return code of 0.

```
$ ping -c 1 192.168.0.1 >/dev/null; echo $?
0
```

But yes, I do think that it is better to inspect return codes than prose. Prose can change at any time. There's absolutely no guarantee of stability.

jyejare: Yeah, Already working on it. Return codes are good to verify. Currently working on timeout thing.

automation_tools/__init__.py
Ichimonji10: > I am seriously not understanding how do I...

I think you missed my point. If ping doesn't provide you with requisite functionality, then don't sit there and bang your head against the wall. I'm simply asking you to read the manual before writing any code.

automation_tools/__init__.py
Ichimonji10: :+1:

jyejare:
@elyezer and @Ichimonji10 , The comments has been considered and the PR has been updated. Please review. Thanks for the comments of Improvement.

automation_tools/__init__.py
```diff
@@ -1622,6 +1623,29 @@ def setenforce(mode):
...
 
+def host_pings(host, attempts=100):
+    """This ensures the given IP/hostname pings succesfully.
+
+    :param host: A string. The IP or hostname of host.
+    :param attempts: An integer.
+        The maximum number of attempts to ping the host.
+
+    """
+    for attempt in range(attempts):
```
elyezer: With this number of attempts we are waiting for 8.333333333 minutes.

I think a better approach here is to get the time stamp and check if 20 minutes have passed since the start of the pooling. then we can expect a timeout in minutes to be passed and defaulting to 20.

elyezer:
ACK, pending comments.

Ichimonji10:
ACK

PR #227, opened 2015-08-24
Updated  install_discovery_image task with setup_foreman_discovery

sghai:
Pre-requisites for foreman-discovery are:
- Image should be installed
- PXE global template needs to be properly configured

I find best and easier way to update the template is via hammer. I placed the template in `sesame` and fetching that via http URL by exporting `PXE_DEFAULT_TEMPLATE_URL` from jenkins.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -445,9 +445,19 @@ def install_puppet_scap_client():
     run('yum -y install puppet-foreman_scap_client', warn_only=True)
 
 
-def install_discovery_image():
-    """Task to install foreman discovery image."""
+def setup_foreman_discovery():
+    """Task to setup foreman discovery."""
+    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
+    template_url = os.environ['PXE_DEFAULT_TEMPLATE_URL']
+    template_file = run('mktemp')
     run('yum -y install foreman-discovery-image', warn_only=True)
+    run('wget -O {0} {1}'.format(template_file, template_url))
+    run('sed -i -e "s/SatelliteCapsule_HOST/{0}/" {1}'
+        .format(env['host'], template_file))
+    run('hammer -u admin -p {0} template update --name '
+        '"PXELinux global default" --file {1}'
+        .format(admin_password, template_file))
+    run('rm -rf {0}'.format(template_file))
 
 
 def vm_create():
@@ -993,7 +1003,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         execute(setup_oscap, host=host)
         execute(katello_service, 'restart', host=host)
         execute(install_puppet_scap_client, host=host)
-        execute(install_discovery_image, host=host)
+        execute(setup_foreman_discovery, host=host)
 
     certificate_url = certificate_url or os.environ.get(
         'FAKE_MANIFEST_CERT_URL')


automation_tools/__init__.py
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
-def install_discovery_image():
-    """Task to install foreman discovery image."""
+def setup_foreman_discovery():
+    """Task to setup foreman discovery."""
+
+    admin_password = os.environ.get('ADMIN_PASSWORD')
+    if admin_password is None:
+        admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
+    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
+    host = env['host']
```
jyejare: Variable 'host' is one time use only variable. You can directly use it instead.
```
run('sed -i -e "s/SatelliteCapsule_URL/{0}/" /var/tmp/template.txt'
''.format(env['host']))
```

jyejare:
ACK pending comment.

automation_tools/__init__.py
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
 
 
-def install_discovery_image():
-    """Task to install foreman discovery image."""
+def setup_foreman_discovery():
+    """Task to setup foreman discovery."""
+
+    admin_password = os.environ.get('ADMIN_PASSWORD')
+    if admin_password is None:
+        admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
```
Ichimonji10: This is kinda wonky. The `get` method is already designed to handle this bit of logic:

> get(key[, default])
>
> > Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.

So the above three lines can be simplified to:

```python
```

Or, if `''` is an acceptable value for `ADMIN_PASSWORD`, do this instead:

try:
    admin_password = os.environ['ADMIN_PASSWORD']
except KeyError:
    admin_password = 'changeme'
```

Also, please drop the blank line between the docstring and the start of the method logic.

File: automation_tools/__init__.py [hunk 3]
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
+    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
```
Ichimonji10: What if this environment variable is not set? Is `None` a legal value? If so, then this is OK. But if this environment variable is required and no sane default can be given, then just perform a regular dict lookup:

```python
template_url = os.environ['PXE_DEFAULT_TEMPLATE_URL']
```

automation_tools/__init__.py
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
+    host = env['host']
+    if template_url is None:
+        print('You should specify the PXE global default template URL')
```
Ichimonji10: How should the user provide this information? No clue is given as to which environment variable should be set. Can we drop this check and perform a regular dict lookup above? If the dict lookup fails, a `KeyError` will be raised, and it will include a reference to the required environment variable.

automation_tools/__init__.py
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
+        sys.exit(1)
     run('yum -y install foreman-discovery-image', warn_only=True)
+    run('wget -O /var/tmp/template.txt {0}'.format(template_url))
+    run('sed -i -e "s/SatelliteCapsule_URL/{0}/" /var/tmp/template.txt'
+        ''.format(host))
```
Ichimonji10: The `''` can be dropped here.

automation_tools/__init__.py
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
+    run('hammer -u admin -p {0} template update --name '
+        '"PXELinux global default" --file /var/tmp/template.txt'
+        ''.format(admin_password))
```
Ichimonji10: The `''` can be dropped here.

automation_tools/__init__.py
```diff
@@ -425,9 +425,24 @@ def setup_scap_client():
...
+    admin_password = os.environ.get('ADMIN_PASSWORD')
+    if admin_password is None:
+        admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
+    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
+    host = env['host']
+    if template_url is None:
+        print('You should specify the PXE global default template URL')
+        sys.exit(1)
     run('yum -y install foreman-discovery-image', warn_only=True)
+    run('wget -O /var/tmp/template.txt {0}'.format(template_url))
```
Ichimonji10: Instead of hard-coding a reference to `/var/tmp/template.txt`, can we use `mktemp` instead?

automation_tools/__init__.py
sghai: +1. will update. thanks

automation_tools/__init__.py
sghai: yeah I think `admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')` would solve the purpose.

automation_tools/__init__.py
sghai: +1 for dict lookup. We are setting the URL env variable via jenkins and `None` is not legal value.

automation_tools/__init__.py
sghai: +1

automation_tools/__init__.py
sghai: yes.. will drop. thanks

sghai:
@all: please re-review this PR.. thanks

automation_tools/__init__.py
```diff
@@ -445,9 +445,19 @@ def install_puppet_scap_client():
     run('yum -y install puppet-foreman_scap_client', warn_only=True)
 
 
-def install_discovery_image():
-    """Task to install foreman discovery image."""
+def setup_foreman_discovery():
+    """Task to setup foreman discovery."""
+    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
+    template_url = os.environ['PXE_DEFAULT_TEMPLATE_URL']
+    template_file = run('mktemp /var/tmp/template.XXXXXXXXXXXX')
```
Ichimonji10: Is there any reason to specifically create a temp file in `/var/tmp` instead of, say, `/tmp`? If there's no special reason to choose one directory over the other, I'd suggest using just `mktemp`, like so:

```
[root@qe-sat6-rhel66 ~]# mktemp
/tmp/tmp.t9hWUeagql
```

sghai: Yeah.. no specific reason. I can use mktemp without using `/var/tmp`

Ichimonji10:
Looks great. ACK.

automation_tools/__init__.py
```diff
@@ -445,9 +445,19 @@ def install_puppet_scap_client():
...
-def install_discovery_image():
-    """Task to install foreman discovery image."""
+def setup_foreman_discovery():
+    """Task to setup foreman discovery."""
+    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
+    template_url = os.environ['PXE_DEFAULT_TEMPLATE_URL']
+    template_file = run('mktemp')
     run('yum -y install foreman-discovery-image', warn_only=True)
+    run('wget -O {0} {1}'.format(template_file, template_url))
+    run('sed -i -e "s/SatelliteCapsule_URL/{0}/" {1}'
```
sthirugn: Should we call `SatelliteCapsule_URL` as `SatelliteCapsule_HOST`?

sghai: Updated.

elyezer:
ACK

PR #232, opened 2015-08-28
Timeout improvements in Upgrade

jyejare:
Timeout improvements in Upgrade Automation!

The number trys/attempts for timeout has been replaced with time in minutes.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1644,15 +1644,16 @@ def setenforce(mode):
     run('setenforce {0}'.format(mode))
 
 
-def host_pings(host, attempts=200):
+def host_pings(host, timeout=15):
     """This ensures the given IP/hostname pings succesfully.
 
     :param host: A string. The IP or hostname of host.
-    :param attempts: An integer.
-        The maximum number of attempts to ping the host.
+    :param timeout: An integer.
+        The timeout in minutes to ping the host.
 
     """
-    for attempt in range(attempts):
+    timeup = time.time() + int(timeout) * 60
+    while True:
         command = subprocess.Popen(
             'ping -c1 {0}; echo $?'.format(host),
             stdout=subprocess.PIPE,
@@ -1662,7 +1663,14 @@ def host_pings(host, attempts=200):
         output = command.communicate()[0]
         # Checking the return code of ping is 0
         if int(output.split()[-1]) == 0:
+            print(
+                'SUCCESS !! The given host {0} has been pinged!!'.format(host))
             break
+        elif time.time() > timeup:
+            print(
+                'The timout for pinging the host {0} has reached!'.format(host)
+            )
+            sys.exit(1)
         else:
             time.sleep(5)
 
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1644,15 +1644,16 @@ def setenforce(mode):
...
 
     :param host: A string. The IP or hostname of host.
-    :param attempts: An integer.
-        The maximum number of attempts to ping the host.
+    :param timeout: An integer.
+        The timeout in minutes to ping the host.
 
     """
-    for attempt in range(attempts):
+    timeup = time.time()+(int(timeout)*60)
```
Ichimonji10: Python's operator precedence is the same as in algebra: multiplication and division before addition and subtraction.

```python
>>> 1 + 2 * 10
21
```

Nitpick: Add some whitespace around each operator and drop unnecessary parentheses:

```python
timeup = time.time() + int(timeout) * 60
```

automation_tools/__init__.py
```diff
@@ -1763,22 +1774,25 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
         network=network.id
     )
     # Assigning floating ip to instance
+    timeup = time.time()+(int(timeout)*60)
```
Ichimonji10: Same nitpick.

automation_tools/__init__.py
```diff
@@ -1763,22 +1774,25 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
...
     while True:
         try:
+            if time.time() > timeup:
+                print('The timeout for assigning the floating IP has reached!')
+                sys.exit(1)
             instance.add_floating_ip(floating_ip)
+            print('SUCCESS!! The flaoting IP has been assigned to instance!')
```
Ichimonji10: "flaoting"

Ichimonji10: Would it be helpful to note information about `floating_ip` or `instance` in these messages?

automation_tools/__init__.py
```diff
@@ -1763,22 +1774,25 @@ def create_openstack_instance(instance_name, image_name, flavor_name, ssh_key):
         network=network.id
     )
     # Assigning floating ip to instance
+    timeup = time.time()+(int(timeout)*60)
     while True:
         try:
+            if time.time() > timeup:
+                print('The timeout for assigning the floating IP has reached!')
+                sys.exit(1)
```
Ichimonji10: Does this need to be in the `try` block? If reasonably possible, only code that might raise a `novaclient.exceptions.BadRequest` should be in the `try` block.

Ichimonji10:
ACK pending comments.

automation_tools/__init__.py
jyejare: +1. Done

automation_tools/__init__.py
jyejare: Done.

automation_tools/__init__.py
jyejare: Noted only Floating IP. The instance name is not required note as we are not using it anywhere to test.

automation_tools/__init__.py
jyejare: +1. Done.

elyezer:
ACK

PR #233, opened 2015-08-30
Add capsule provisioning tasks

### LINKED ISSUE #185: Capsules: Implement parent-side (satellite) prerequisites for setting up external capsule
*Complement to #186* 

In preparation for being able to automate capsule installation/configuration, there are bits that need to be done on the satellite side of the house.  There may or may not be some of this in automation-tools already; if there are, we'll simply need to be able to tweak params.

REQUIREMENTS
* Ability to install multiple capsules in serial or parallel
* Ability to sync multiple capsules in serial or parallel (preferably async)
* Ability to halt process if something goes wrong (i.e., sync not triggered if a capsule install/configuration fails)

Steps
* sync capsule content
* create content view with RHEL+Tools+capsule (RHEL7)
* publish/promote content view
* create activation key associated with content view
* generate capsule cert 

Proposed automation-tools namespace [and possible parameters/inputs]

* **capsule_sync** *[URL_for_capsule_content]*
* **capsule_create_cv**
* **capsule_publish_cv**
* **capsule_ak_create** 
* **capsule_generate_cert**

### LINKED ISSUE #186: Capsules: Implement child-side (capsule) prerequisites for setting up external capsules
*Complement to #185* 

Implement the child-side (capsule) functionality in automation-tools that will allow us to complete a capsule install.  There may be some functionality in here that can piggyback off existing automation-tools.

* copy capsule cert to child
* register target capsule(s) using content view
  * retrieve sat ca cert
  * register
* install rpms on child 
* install product via capsule-installer 
* partition_disk
* trigger sync on all content

Proposed namespace [and possible parameters]
* **capsule_copy_cert** *[cert name, child_server]*
* **capsule_register** *[URL_for_ca_cert_, parent_server, child_server]*
* **capsule_install_rpms**
* **capsule_installer**
* **capsule_sync**

elyezer:
The capsule module contains helper tasks in order to accomplish the
capsule registering automation. They rely on a json configuration and
expect some data to be synced and set on Satellite before running it.

This is basically katello-rapid node migrated to be used inside
automation-tools.

Close #185 and close #186.

### DIFF
--- a/automation_tools/satellite6/capsule.py
+++ b/automation_tools/satellite6/capsule.py
@@ -1,0 +1,399 @@
+"""Tasks for helping automating the provisioning of Satellite 6 Capsules"""
+from __future__ import print_function
+import json
+
+from fabric.api import env, get, put, run, settings, task
+from fabric.operations import _AttributeList
+
+
+class _AttributeDict(dict):
+    """Simple dict subclass to allow arbitrary attibute access"""
+
+
+class ImproperlyConfigured(Exception):
+    """Indicates that the capsule configuration is somehow improperly
+    configured
+    - for example, if the configuration is not loaded or some required
+    configuration is missing.
+
+    """
+
+
+class Credentials(object):
+    """Stores a server SSH credentials information.
+
+    Usage::
+
+        cred1 = Credentials('admin', 'password')
+        cred2 = Credentials(key_filename='/path/to/ssh.key')
+
+    """
+    def __init__(self, user=None, password=None, key_filename=None):
+        self.user = user
+        self.password = password
+        self.key_filename = key_filename
+
+
+class HostConfig(Credentials):
+    """Stores a host's hostname and credentials information.
+
+    Usage::
+
+        cred1 = HostConfig('host1.example.com', 'admin', 'password')
+        cred2 = HostConfig(
+            'host2.example.com', key_filename='/path/to/ssh.key')
+
+    """
+    def __init__(self, hostname=None, port=22, *args, **kwargs):
+        super(HostConfig, self).__init__(*args, **kwargs)
+        self.hostname = hostname
+        self.port = port
+
+    @property
+    def host_string(self):
+        """Return a host_string in the format expected by Fabric"""
+        return '{}@{}:{}'.format(self.user, self.hostname, self.port)
+
+
+class Config(object):
+    """Configuration information provide easy access to configuration and some
+    helper methods to identify if some configuration is present or not.
+
+    """
+    def __init__(self, path):
+        self.path = path
+        self.organization_label = None
+        self.environment = None
+        self.content_view = None
+        self.activation_key = None
+        self.admin_user = None
+        self.admin_password = None
+        self.defaults = None
+        self.server = None
+        self.capsules = []
+        self._key_filenames = set()
+
+        self._parse()
+
+    def _parse(self):
+        """Parse the configuration and store the contents"""
+        with open(self.path) as handler:
+            data = json.load(handler)
+        self.organization_label = data.get('organization-label')
+        self.environment = data.get('environment')
+        self.content_view = data.get('content-view')
+        self.activation_key = data.get('activation-key')
+        self.admin_user = data.get('admin-user')
+        self.admin_password = data.get('admin-password')
+        defaults = data.get('defaults')
+        if defaults is not None and isinstance(defaults, dict):
+            key_filename = defaults.get('key-filename')
+            self._key_filenames.add(key_filename)
+            self.defaults = Credentials(
+                user=defaults.get('user'),
+                password=defaults.get('password'),
+                key_filename=key_filename,
+            )
+        server = data.get('server')
+        if server is not None and isinstance(server, dict):
+            key_filename = server.get('key-filename')
+            self._key_filenames.add(key_filename)
+            self.server = HostConfig(
+                hostname=server.get('hostname'),
+                user=server.get('user', self.defaults.user),
+                password=server.get('password', self.defaults.password),
+                key_filename=key_filename,
+            )
+        capsules = data.get('capsules')
+        if capsules is not None and isinstance(capsules, list):
+            for capsule in capsules:
+                if capsule is not None and isinstance(capsule, dict):
+                    key_filename = capsule.get('key-filename')
+                    self._key_filenames.add(key_filename)
+                    self.capsules.append(HostConfig(
+                        hostname=capsule.get('hostname'),
+                        user=capsule.get('user', self.defaults.user),
+                        password=capsule.get(
+                            'password', self.defaults.password),
+                        key_filename=key_filename,
+                    ))
+
+    @property
+    def key_filenames(self):
+        """Return a list of collect key filenames or None if the list is
+        empty.
+
+        """
+        if self._key_filenames:
+            return list(self._key_filenames)
+        else:
+            return None
+
+    @property
+    def passwords(self):
+        """Return a dict in the format suited for Fabric usage in order to
+        define passwords for hosts.
+
+        """
+        passwords = {}
+        if self.server.password and not self.server.key_filename:
+            passwords[self.server.host_string] = self.server.password
+        for capsule in self.capsules:
+            if capsule.password and not capsule.key_filename:
+                passwords[capsule.host_string] = capsule.password
+        return passwords
+
+
+def _get_config():
+    """Get the capsule configuration if available in the fabric environment
+    else raise ``ImproperlyConfigured``.
+
+    """
+    config = env.get('capsule_config')
+    if config is None:
+        raise ImproperlyConfigured(
+            'Make sure to run load_capsule_config task.')
+    return config
+
+
+def _lower_dict_keys(d):
+    """Helper for ensuring that all dictionary keys are lowercase."""
+    if isinstance(d, list):
+        return [_lower_dict_keys(v) for v in d]
+    elif isinstance(d, dict):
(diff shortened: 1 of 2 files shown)


automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+from fabric.api import env, get, put, run, settings, task
+from fabric.operations import _AttributeList
+
+
+class _AttributeDict(dict):
+    """Simple dict subclass to allow arbitrary attibute access"""
+
+
+class ImproperlyConfigured(Exception):
+    """Indicates that the capsule configuration is somehow improrpely configure
```
Ichimonji10: "improperly configured"

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+            key_filename = server.get('key-filename')
+            self._key_filenames.add(key_filename)
+            self.server = HostConfig(
+                hostname=server.get('hostname'),
+                user=server.get('user', self.defaults.user),
+                password=server.get('password', self.defaults.password),
+                key_filename=key_filename,
+            )
+        capsules = data.get('capsules')
+        if capsules is not None and isinstance(capsules, list):
```
Ichimonji10: Does it make sense to do the `isinstance` check here? Knowing that a non-default value has been passed in is enough to trigger parsing logic. Checking for a specific class prevents you from taking advantage of Python's duck typing support.

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+        if defaults is not None and isinstance(defaults, dict):
+            key_filename = defaults.get('key-filename')
+            self._key_filenames.add(key_filename)
+            self.defaults = Credentials(
+                user=defaults.get('user'),
+                password=defaults.get('password'),
+                key_filename=key_filename,
+            )
+        server = data.get('server')
+        if server is not None and isinstance(server, dict):
```
Ichimonji10: Same question about duck typing here.

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+        cred2 = HostConfig(
+            'host2.example.com', key_filename='/path/to/ssh.key')
+
+    """
+    def __init__(self, hostname=None, port=22, *args, **kwargs):
+        super(HostConfig, self).__init__(*args, **kwargs)
+        self.hostname = hostname
+        self.port = port
+
+    @property
```
Ichimonji10: Looks reasonable. Is it correct to think that setters or other additional logic may be implemented in the future?

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+                        user=capsule.get('user', self.defaults.user),
+                        password=capsule.get(
+                            'password', self.defaults.password),
+                        key_filename=key_filename,
+                    ))
+
+    @property
+    def key_filenames(self):
+        """Return a list of collect key filenames or None if the list is
+        empty.
```
Ichimonji10: I don't see why `None` is returned. In the few cases where this property is used boolean type casting is used:

```python
        if self.server.password and not self.server.key_filename:
            passwords[self.server.host_string] = self.server.password
        for capsule in self.capsules:
            if capsule.password and not capsule.key_filename:
```

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+def _get_config():
+    """Get the capsule configuration if available in the fabric environment
+    else raise ``ImproperlyConfigured``.
+
+    """
+    config = env.get('capsule_config')
+    if config is None:
+        raise ImproperlyConfigured(
+            'Make sure to run load_capsule_config task.')
+    return config
```
Ichimonji10: Could this be simplified by asking for forgiveness?

```python
try:
    return env['capsule_config']
except KeyError as err:
    raise ImproperlyConfigured(
        'Make sure to run the load_capsule_config task: {}'
        .format(err)
    )
```

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+
+    @property
+    def host_string(self):
+        """Return a host_string in the format expected by Fabric"""
+        return '{}@{}:{}'.format(self.user, self.hostname, self.port)
+
+
+class Config(object):
+    """Configuration information provide easy access to configuration and some
+    helper methods to identify is some configuration is present or not.
```
sthirugn: s/is/if

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+def sync_capsule_content(capsule):
+    """Start content synchronization in the capsule. The content
+    synchronization will be asynchronously, check the capsule logs to see when
+    it have finished.
+
+    :param dict capsule: A capsule dictionary containing its ``id`` and
+        ``name``.
+
+    """
+    if capsule['id'] != 1:
```
Ichimonji10:
```python
if capsule['id'] == 1:
    print('Skipping default capsule...')
# remainder of logic is at a lower indentation level
```

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+    elif isinstance(result, dict):
+        result = _AttributeDict(result)
+    result.succeeded = command_result.succeeded
+    result.failed = command_result.failed
+    result.return_code = command_result.return_code
+    return result
+
+
+@task
+def hammer_capsule_lcenvs(capsule_id):
```
sthirugn: will the name get_capsule_lcenvs make more sense?

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+@task
+def hammer_capsule_add_lcenv(capsule_id, lcenv_id):
+    """Add the lifecycle environment to the capsule.
+
+    :param capsule_id: The capsule ID to add the lifecycle environment.
+    :param lcenv_id: The lifecycle environment ID to add to the capsule.
+
+    """
+    return hammer(
+        'capsule content add-lifecycle-environment --environment-id {} --id {}'
```
sthirugn: are we not emphasizing the use of numbers in {} anymore?

automation_tools/satellite6/capsule.py
```diff
@@ -0,0 +1,398 @@
...
+
+    """
+    return hammer(
+        'capsule content available-lifecycle-environments --id {}'
+        .format(capsule_id),
+    )
+
+
+@task
+def hammer_capsule_add_lcenv(capsule_id, lcenv_id):
```
sthirugn: will the name add_capsule_lcenv make more sense?

automation_tools/satellite6/capsule.py
elyezer: The later code assumes that it have a list on capsules, if a string is received, for example, I will do a for over its chars.

With that said I decided to ensure the expected type and just ignore other types.

automation_tools/satellite6/capsule.py
elyezer: Same answer :smile:

automation_tools/satellite6/capsule.py
elyezer: For this case just the getter. It is just a helper to generate the expected format for the host_string.

automation_tools/satellite6/capsule.py
elyezer: will update

automation_tools/satellite6/capsule.py
elyezer: None is the default value for that config on Fabric, so this method is ensuring that.

automation_tools/satellite6/capsule.py
elyezer: Good approach, but will leave as is because there are other places in automation-tools that uses similar approach.

automation_tools/satellite6/capsule.py
elyezer: Maybe. The current name was choose because this is a wrapper around the hammer task. With that said was decided to use hammer prefix instead of get.

I don't have a personal preference here, I can update if that will really make more sense.

automation_tools/satellite6/capsule.py
elyezer: Same as previous, was named this way to give information about that is running hammer and performing the action of "capsule add-lifecycle-environment".

As before I don't have a personal preference and I can update this.

automation_tools/satellite6/capsule.py
elyezer: That is just to support python 2.6, I don't see a real benefit of supporting that version. The numbering is required only if that makes sense, otherwise it will use the order of the arguments passed to format, so less typing.

automation_tools/satellite6/capsule.py
Ichimonji10: Hmm. I suspect there's a better way to handle this. But your explanation is plausible and I don't have a better solution in hand, so OK.

sthirugn:
ACK

automation_tools/satellite6/capsule.py
Ichimonji10: Ahh. This makes more sense. Thanks for explaining.

automation_tools/satellite6/capsule.py
Ichimonji10: This has no side effects, so it seems OK.

Ichimonji10:
ACK

PR #235, opened 2015-08-31
Removed discovery image install for upstream

sghai:
Since we don't have package for discovery image in upstream. So this task is not applicable for upstream.

Thanks @kbidarkar for pointing this.


sthirugn:
ACK

Ichimonji10:
ACK

PR #237, opened 2015-09-16
Updating pxe template with hostname instead of IP

sghai:
I realized on our automation machines template is being updated with IP of server. Ideally it should be updated with hostname.


Ichimonji10:
`env['host']` appears in four other locations in the same module. Should those other occurences also be updated?

sghai:
I don't think we need to update other locations because we need this change only for `setup_foreman_discovery`

Ichimonji10:
OK. ACK

sthirugn:
ACK

PR #239, opened 2015-09-21
Ignore errors while fetching the Virtual Bridge information.

kbidarkar:
* Errors are encountered when the physical hardware has infiniband
   devices.
* We just need virtual bridge information at this stage.


elyezer:
ACK

omaciel:
ACK

PR #240, opened 2015-09-24
Conditionally setups Foreman Discovery

elyezer:
Run setup_foreman_discovery on product_install only if the required
environment variable is defined.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -452,7 +452,11 @@ def install_puppet_scap_client():
 def setup_foreman_discovery():
     """Task to setup foreman discovery."""
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
-    template_url = os.environ['PXE_DEFAULT_TEMPLATE_URL']
+    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
+    if template_url is None:
+        print('You must specify the PXE default template URL')
+        sys.exit(1)
+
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -452,7 +452,11 @@ def install_puppet_scap_client():
 def setup_foreman_discovery():
     """Task to setup foreman discovery."""
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
-    template_url = os.environ['PXE_DEFAULT_TEMPLATE_URL']
+    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
+    if template_url is None:
+        print('You must specify the PXE default template URL')
+        sys.exit(1)
```
omaciel: Will this exit the entire script? Should it only `return` instead?

omaciel:
ACK pending comment

automation_tools/__init__.py
elyezer: This is the current solution when failing for a missing env var. Other tasks use same approach.

This is not the preferred way when using as a library, but it is really hard to work with fabric as a library since we need to to everything to not allow fabric exiting the interpreter.

This check if more for a standalone call than for product_install since it checks if the env var is define in order to call this task.

sghai:
:+1:  Thanks Elyezer

oshtaier:
Thanks @elyezer . I will try that fix right away

PR #241, opened 2015-09-29
Added support to provision Sat6 VMs using static ips & VLAN Bridges

kbidarkar:
*) Now setup_ddns will run only when VM uses a dhcp IP via br0
   Bridge.
*) Dynamic DNS is not required when we provide the VM with a static
   Ip address via VLAN Bridges.
*) The idea is to ask the DNS configured beforehand.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -489,9 +489,28 @@ def vm_create():
         path where the generated image will be stored
     CPU_FEATURE
         copies cpu features of base-metal to vm, thus enabling nested_virt
+    BRIDGE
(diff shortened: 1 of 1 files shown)


sghai:
Changes Looks good. Assuming VM's Hostname and static IP binding is done.

kbidarkar:
Yup.
VM's Hostnames have been added to IPA server which exists in RDULab.
Static IP binding is done in Jenkins --> provisioning Config File

sghai:
:+1:

elyezer:
Looks good. ACK

PR #249, opened 2015-10-07
A couple of tweaks and a fix.

omaciel:
* Fixed 'delete_custom_repos' function signature.
* Delete EPEL repositories for RHEL 6.x configurations on
  'setup_default_docker' task.
* Switched 'debug' flag for 'katello-installer' to 'False' so that
  answers file is generated and kept.


Ichimonji10:
ACK

elyezer:
ACK

PR #251, opened 2015-10-13
generate_capsule_certs method to force create the capsule cert

jyejare:
Force Generate the capsule cert even if present/generated already.

### DIFF
--- a/automation_tools/satellite6/capsule.py
+++ b/automation_tools/satellite6/capsule.py
@@ -197,8 +197,8 @@ def get_oauth_info():
 
 
 @task
-def generate_capsule_certs(capsule_hostname):
-    """Generate certificates for a capusule.
+def generate_capsule_certs(capsule_hostname, force=False):
+    """Generate certificates for a capsule.
 
     Run ``capsule-certs-generate --capsule-fqdn <capsule_hostname> --certs-tar
     "<capsule_hostname>-certs.tar"`` in order to generate them.
@@ -208,11 +208,13 @@ def generate_capsule_certs(capsule_hostname):
 
     :param str capsule_hostname: The fully qualified domain name for the
         capsule.
+    :param bool force: Force creation of the capsule cert even if it is
+        already created.
 
     """
     cert_path = '{}-certs.tar'.format(capsule_hostname)
     result = run('[ -f {} ]'.format(cert_path), quiet=True)
-    if result.failed:
+    if result.failed or force:
         run('capsule-certs-generate -v --capsule-fqdn {} '
             '--certs-tar {}'.format(capsule_hostname, cert_path))
     return cert_path


automation_tools/satellite6/capsule.py
```diff
@@ -212,7 +212,7 @@ def generate_capsule_certs(capsule_hostname):
     """
     cert_path = '{}-certs.tar'.format(capsule_hostname)
     result = run('[ -f {} ]'.format(cert_path), quiet=True)
-    if result.failed:
+    if result.failed or force:
         run('capsule-certs-generate -v --capsule-fqdn {} '
             '--certs-tar {}'.format(capsule_hostname, cert_path))
```
rplevka: Shouldn't we also check on the result of this command before returning the `cert_path`?
what if it somehow fails?

rplevka:
Update the docstring of the method with:
```:param bool force: <parameter description>```

File: automation_tools/satellite6/capsule.py [hunk 2]
```diff
@@ -208,14 +208,19 @@ def generate_capsule_certs(capsule_hostname):
...
-    return cert_path
+    if int(run('[ -f {} ]; echo $?'.format(cert_path))) == 0:
+        return cert_path
+    else:
+        print('Warning ! Failed to generate Capsule Cert on Satellite !')
```
rplevka: This should be probably replaced by a proper exception raise.
Not a blocker

rplevka:
thanks for addressing my comments.
ACK

automation_tools/satellite6/capsule.py
jyejare: We are not raising the exceptions in automation_tools. Instead we are just printing the impacts of failure and proceeding fwd.

automation_tools/satellite6/capsule.py
```diff
@@ -208,14 +208,19 @@ def generate_capsule_certs(capsule_hostname):
 
     :param str capsule_hostname: The fully qualified domain name for the
         capsule.
+    :param bool force: To force create the capsule cert if already created.
+    By default wont force create if already created.
```
elyezer: Please ident this line.

automation_tools/satellite6/capsule.py
elyezer: There is no need to check if the file is created, if the capsule-certs-generate fails it will stop the execution.

elyezer:
Sorry for the long delay to review this.

ACK, pending comments.

jyejare:
@elyezer , worked on your comment, please have a second look on the PR.

automation_tools/satellite6/capsule.py
```diff
@@ -208,11 +208,13 @@ def generate_capsule_certs(capsule_hostname):
 
     :param str capsule_hostname: The fully qualified domain name for the
         capsule.
+    :param bool force: Force creation of the capsule cert even if it is
+    already created.
```
elyezer: Please ident this line. See capsule_hostname param as an example.

elyezer: Like this:

```diff
+    :param bool force: Force creation of the capsule cert even if it is
+        already created.
```

sthirugn: @jyejare if you like -  on your robottelo root folder, run 'make docs' and inspect the htmls to see why @elyezer is asking for indentation.

PR #252, opened 2015-10-15
Force Create the Capsule cert in Capsule upgrade

jyejare:
Force Create the Capsule Cert while doing upgrade because already there is one used to setup the capsule. And hence for upgrading the capsule we need to create one more. So force creating capsule cert even if already present.


elyezer:
ACK

rplevka:
ACK

PR #253, opened 2015-10-27
Fix for prompt of Host Key checking

jyejare:
This fixes the prompt for yes/no while Host key checking when doing scp.


oshtaier:
ACK

abalakh:
ACK

PR #257, opened 2015-11-09
Missed word s and return code fix

jyejare:
I missed letter 's' in file name 'authorized_keys', triggers the failures in automation.
And also the fix for return code 1 with 'warn_only = True'.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1756,8 +1756,10 @@ def copy_ssh_key(from_host, to_host):
     :param to_host: A string. Hostname on to which the ssh-key will be copied.
 
     """
-    execute(lambda: run('[ ! -f ~/.ssh/id_rsa.pub ] && ssh-keygen '
-                        '-f ~/.ssh/id_rsa -t rsa -N \'\''), host=from_host)
+    execute(lambda: run(
+        '[ ! -f ~/.ssh/id_rsa.pub ] && '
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1756,8 +1756,10 @@ def copy_ssh_key(from_host, to_host):
     :param to_host: A string. Hostname on to which the ssh-key will be copied.
 
     """
-    execute(lambda: run('[ ! -f ~/.ssh/id_rsa.pub ] && ssh-keygen '
-                        '-f ~/.ssh/id_rsa -t rsa -N \'\''), host=from_host)
+    execute(lambda: run(
+        '[ ! -f ~/.ssh/id_rsa.pub ] && '
+        'ssh-keygen -f ~/.ssh/id_rsa -t rsa -N \'\'', warn_only=True),
```
elyezer: This should not fail if was not able to generate the key? Adding warn_only will allow the code proceed without this prerequisite.

elyezer:
ACK, pending question.

automation_tools/__init__.py
jyejare: Anyhow I am checking below whether the id_rsa.pub is generated or not.
But if It is already there it returns code 1 which causes fabric to terminate the execution. Saying the return code is 1.

elyezer: Talked with @jyejare and the warn_only is needed to ensure that fabric will not stop the execution if the private key since the checking will return false (non-zero). Anyway there is another check later to ensure the private key exists.

PR #260, opened 2015-11-11
Added functions idp_authenticate and download_manifest

rplevka:
-these functions are designed for downloading the manifest file
    directly from the customer portal


rplevka:
`idp_authenticate()` works locally.
It takes the credentials and IDP url and mimic the browser authentication ping-pong.
After authentication is successful, it spit out a `requets.session` object.

This object is used by the other function - `download_manifest()`.
Besides a `session` object, this one accepts also `URL` of the customer portal (such as access.redhat.com)
and the `distributor hash`, identifying the target satellite registration.
The function extracts the session cookie from the `session` object, constructs a `curl` command and executes it remotly on the target machine.
The output of the function is a path to a downloaded manifest file.

automation_tools/__init__.py
```diff
@@ -7,11 +7,13 @@
 from __future__ import print_function
 import os
 import random
+import requests
```
elyezer: This is not part of requirements. Need to add there.

automation_tools/__init__.py
```diff
@@ -7,11 +7,13 @@
...
 import socket
 import sys
 import time
 import novaclient
 import subprocess
+from lxml import html
```
elyezer: I think we can use some xml parsing module from stdlib and avoid adding another dependency.

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,133 @@ def product_upgrade(
...
+def idp_authenticate(idp=None):
+    """Authenticates on a customer portal and returns a session object.
+
+    :param idp: id provider url (typically https://idp.redhat.com/idp/)
+
+    The following environment variables affect this command:
+
+    IDP_URL
+        Identity Provider URL
+
```
elyezer: According to PEP8, this blank line is not necessary.

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,133 @@ def product_upgrade(
         execute(satellite6_capsule_upgrade, host=cap_host)
         # Generate foreman debug on capsule
         execute(foreman_debug, 'capsule', host=cap_host)
+
+
+def idp_authenticate(idp=None):
+    """Authenticates on a customer portal and returns a session object.
+
+    :param idp: id provider url (typically https://idp.redhat.com/idp/)
```
elyezer: Move param definition to bottom of the docstring.

automation_tools/__init__.py
elyezer: I recommend using ElementTree.

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,133 @@ def product_upgrade(
...
+        return session
+    else:
+        auth_request.raise_for_status()
+
+
+def download_manifest(url=None, session=None, distributor=None):
+    """Task for downloading the manifest file from customer portal.
+    Returns a path to a downloaded manifest file
+
+    :url: Customer Portal URL (typically 'https://access.redhat.com')
```
elyezer: Move parameters documentation to bottom of the docstring.

automation_tools/__init__.py
rplevka: @elyezer despite the security vulnerabilites?
https://docs.python.org/2/library/xml.html#xml-vulnerabilities

elyezer: @rplevka if you really doubt the system you are parsing then you should worry about an attack. The safest way, though, is to use an API then even try to parse html/xml.

rplevka: @elyezer ok. Anyway, xml.Etree has trouble parsing the document. Looks like it is malformed in its eyes. I vote for staying with lxml.

rplevka:
- added `requests` to requirements.txt
- addressed comments regarding docstrings

- still relying on `lxml` module

elyezer:
@rplevka requests and lxml should be listed on setup.py file under the install_requires https://github.com/SatelliteQE/automation-tools/blob/master/setup.py#L24

elyezer:
ACK, pending comment.

rplevka:
@elyezer: comments addressed:

modified line 24:
```python
install_requires=['Fabric', 'lxml', 'python-novaclient', 'requests'],
```

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,129 @@ def product_upgrade(
...
         execute(foreman_debug, 'capsule', host=cap_host)
+
+
+def idp_authenticate(idp=None):
+    """Authenticates on a customer portal and returns a session object.
+
+    The following environment variables affect this command:
+
+    IDP_URL
+        Identity Provider URL
```
sthirugn: Do you want to add RHN_USERNAME and RHN_PASSWORD here?

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,129 @@ def product_upgrade(
...
+    :param idp: id provider url (typically https://idp.redhat.com/idp/)
+
```
sthirugn: can you add the return object info here as well?

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,129 @@ def product_upgrade(
...
+    Returns a path to a downloaded manifest file
+
+    The following environment variables affect this command:
+
+    CUSTOMER_PORTAL_URL
+        Customer Portal URL (typically 'https://access.redhat.com')
+    DISTRIBUTOR
+        A distributor hash to be used for getting the manifest
+
+    :url: Customer Portal URL (typically 'https://access.redhat.com')
```
sthirugn: please use the other format like `:param idp:`

automation_tools/__init__.py
```diff
@@ -2283,3 +2285,129 @@ def product_upgrade(
...
+    :session: A requests.session object with a valid customer portal session
+    :distributor: A distributor hash to be used for getting the manifest
```
sthirugn: document the return object as well

sthirugn:
ACK pending nitpicks

rplevka:
@sthirugn 
Thanks for a review. Comments addressed

requirements.txt
```diff
@@ -5,3 +5,5 @@
 # See: https://caremad.io/2013/07/setup-vs-requirement/
 
 --editable .
+
+requests
```
elyezer: since requests is already on setup.py there is no need to modify the requirements.

elyezer:
ACK, pending last comment.

elyezer:
APT

elyezer:
Once Travis is OK I will merge this PR. Further improvements will be handled as needed.

PR #264, opened 2015-11-24
Task to setup private key on satellite server to communicate with libvirt hypervisor

sghai:


sghai:
I've already generated a key pair and placed it in sesame server. We will wget the private key and place it on all Sat servers via this task, so that each satellite can communicate with hypervisor in a secure way and with same key every time.

sghai:
This will cover the 2) point in https://github.com/SatelliteQE/robottelo/issues/3063

automation_tools/__init__.py
```diff
@@ -479,6 +479,16 @@ def setup_foreman_discovery():
     run('rm -rf {0}'.format(template_file))
 
 
+def setup_libvirt_key():
+    """Task to setup key pairs for secure comminication between Satellite server
+    and libvirt hypervisor."""
+    key_url = os.environ.get('LIBVIRT_KEY_URL')
```
kbidarkar: We do have the corresponding entry updated in jenkins configuration file: `Installation Environment Variables`

kbidarkar: :+1:

kbidarkar:
ACK

elyezer:
ACK

PR #265, opened 2015-11-25
Updated private key permissions explicitly

sghai:
The change in key permissions is required.  When we `wget` the key from remote server, the file permissions changed to `644`. However private key should have `600`. So updating the key permissions here with `chmod`.


kbidarkar:
ACK

abalakh:
ACK

Issue #270, opened 2015-12-04
katello installer should run with dhcp-range option while setting up default capsule


sghai:
Currently, we are running katello-installer without `--capsule-dhcp-range` option. However it is required for host discovery. Without this option, a pxe boot host won't get IP from dhcp server.

I think setting environment variable `dhcp_range` in jenkins -> Provisioning Env variables should solve the issue.

kbidarkar:
The PR is referenced above and we now also have the same support for even satellite6-installer robottelo-ci#110

PR #271, opened 2015-12-04
updated installer with dhcp-range option

sghai:
`DHCP_RANGE` environment variable is defined with some specific range in jenkins.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -250,6 +250,11 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         print('Was not possible to fetch domain information')
         sys.exit(1)
 
+    dhcp_range = os.environ.get('DHCP_RANGE')
+    if dhcp_range is None:
+        print('Could not fetch the dhcp range information')
+        sys.exit(1)
+
     if interface is None:
         run('yum install -y libvirt')
         manage_daemon('enable', 'libvirtd')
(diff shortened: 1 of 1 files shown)


sghai:
Covers #270

sghai:
@kbidarkar : Please take a look and suggest if any changes are required.

automation_tools/__init__.py
```diff
@@ -250,6 +250,11 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         print('Was not possible to fetch domain information')
         sys.exit(1)
 
+    dhcp_range = os.environ.get('DHCP_RANGE')
```
kbidarkar: Ok, so DHCP_RANGE is updated with the valid ranges.

kbidarkar:
ACK

sthirugn:
ACK

automation_tools/__init__.py
omaciel: Wait. so we're not required to provide it? What if I don't want to do it? Couldn't we make it optional? << @kbidarkar @sghai

sthirugn:
@omaciel I didnt notice your comment before merging, sorry about that.

omaciel:
@sghai would you mind sending another PR which would make providing the DHCP_RANGE optional?

elyezer:
@omaciel @sthirugn @sghai created #272 to make it optional.

kbidarkar:
Note to self, while reviewing, I should also make a point to read/view the entire code. Looking at the changes it looks good but may be altering the code behavior.

kbidarkar:
:+1:  @omaciel, @elyezer  for finding it and the fix. :)

PR #279, opened 2016-01-06
Added support for configuring manifest links of the downloaded manifest

kbidarkar:


automation_tools/__init__.py
```diff
@@ -2532,3 +2538,20 @@ def download_manifest(url=None, session=None, distributor=None):
...
+
+
+def configure_manifest(manifest=None, prefix_dir=None):
+    """Configures the latest downloaded manifest file to the manifest_latest.zip
+    softlink.
+
+    :param manifest: Specify the manifest file path.
+    :param prefix_dir: The prefix directory to create the temp file.
+    """
+    if manifest:
```
elyezer: The following is redundant

```
if manifest:
    manifest = manifest
```

Just check for other cases.

automation_tools/__init__.py
```diff
@@ -2532,3 +2538,20 @@ def download_manifest(url=None, session=None, distributor=None):
...
+        manifest = manifest
+    elif (prefix_dir is None and manifest is None):
+        manifest = download_manifest(prefix_dir='/opt/manifests/')
+    else:
+        manifest = download_manifest(prefix_dir=prefix_dir)
+    run('unlink /opt/manifests/manifest-latest.zip')
```
elyezer: If I specify a prefix, that means that it should be used here too, isn't it?

automation_tools/__init__.py
```diff
@@ -2532,3 +2538,20 @@ def download_manifest(url=None, session=None, distributor=None):
...
+    run('ln -s {0} /opt/manifests/manifest-latest.zip'.format(manifest))
```
elyezer: Should be used the prefix?

kbidarkar:
I at times re-factor the code at last moment and it becomes funny looking code. :)

rplevka:
@kbidarkar thanks for addressing our comments,
ACK

kbidarkar:
So here is the tested code and it was able to create the manifest on the desired server.
```
[xyz@abc.redhat.com] run: mv /tmp/tmp.bEPLLWgp5m.zip /opt/manifests/
[xyz@abc.redhat.com] run: chmod 644 /opt/manifests/tmp.bEPLLWgp5m.zip
[xyz@abc.redhat.com] run: unlink /opt/manifests/manifest-latest.zip
[xyz@abc.redhat.com] run: ln -s /opt/manifests/tmp.bEPLLWgp5m.zip /opt/manifests/manifest-latest.zip

Done.
Disconnecting from xyz@abc.redhat.com... done.
```

sthirugn:
ACK

elyezer:
ACK

omaciel:
Just wondering if we should perform some type of assertion for `manifest_file = download_manifest()` to make sure that something valid was returned before moving and unlinking files around?

rplevka:
@omaciel 
`download_manifest()` returns the attachment (in case the return header contains that):
https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L2553
Otherwise it raises a `ValueError`.

omaciel:
Thanks for the update @rplevka

PR #280, opened 2016-01-07
 Added support for using custom interface for DHCP and DNS features

kbidarkar:
*) We need support to configure katello-installer if the interface
    is other than eth0, for this I have added support to specify
    the interface in robottelo-ci for sat6-installer job.
    
*) For configuring DHCP and DNS Capsule feature this is required.
    
*) This will mostly be useful for installing Satellite6 in vmware
    where in the network interface is not a standard eth0. Helpful
    during compute resource testing with VmWare.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1064,21 +1064,26 @@ def product_install(distribution, create_vm=False, certificate_url=None,
 
     if distribution in (
             'satellite6-cdn', 'satellite6-downstream', 'satellite6-iso'):
-        # Execute with 'interface=eth0' only if using VLAN Bridges.
-        if os.environ.get('BRIDGE') is None:
-            # execute returns a dictionary mapping host strings to the given
-            # task's return value.
-            installer_options.update(execute(
-                setup_default_capsule, host=host, run_katello_installer=False
-            )[host])
-        else:
+        # When using VLAN Bridges os.environ.get('BRIDGE') is 'true' and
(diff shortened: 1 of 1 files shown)


sthirugn:
APT

automation_tools/__init__.py
```diff
@@ -1071,6 +1071,15 @@ def product_install(distribution, create_vm=False, certificate_url=None,
             installer_options.update(execute(
                 setup_default_capsule, host=host, run_katello_installer=False
             )[host])
+        elif os.environ.get('INTERFACE'):
```
elyezer: This code can be written like this:

```python
          else:
              # execute returns a dictionary mapping host strings to the given
              # task's return value
             installer_options.update(execute(
                 setup_default_capsule,
                 host=host,
                 interface=os.environ.get('INTERFACE') or 'eth0',
                 run_katello_installer=False
             )[host])
```

I don't used os.environ.get('INTERFACE', 'eth0') because I want to handle `None` and `''` will default to `'eth0'`.

kbidarkar: I think this would be a problem if `os.environ.get('INTERFACE') would be TRUE , but it would never enter the `else:` part of the loop as we are checking for `os.environ.get('BRIDGE') is None`.

So I updated it accordingly by taking some hints from your suggestion.

elyezer:
ACK

sghai:
ACK

jyejare:
Great Addition ! ACK !

PR #285, opened 2016-02-09
Runs katello-installer in debug mode.

kbidarkar:
a) katello-installer runs in debug mode
b) This will also cause the parameters being passed to
   katello-installer to not get saved to answers-file.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1725,6 +1725,8 @@ def katello_installer(debug=False, sam=False, verbose=True, **kwargs):
             extra_options.append(
                 '--capsule-dns-forwarders="{0}"'.format(forwarder))
 
+    debug = (debug or os.environ.get('DEBUG', '') == 'true')
+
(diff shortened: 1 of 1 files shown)


kbidarkar:
Fixes issue https://github.com/SatelliteQE/robottelo-ci/issues/37

automation_tools/__init__.py
```diff
@@ -1725,6 +1725,8 @@ def katello_installer(debug=False, sam=False, verbose=True, **kwargs):
             extra_options.append(
                 '--capsule-dns-forwarders="{0}"'.format(forwarder))
 
+    debug = (debug or os.environ.get('DEBUG', '') == 'true')
```
elyezer: Maybe product_install is the better place to put this. I mean, add the logic to get the environment variable and pass the value when It calls katello_installer.

This is a helper function and there is not environment variable processing in here. Because that I feel that we should leave it as a helper function and avoid the introduction of environement variables leaving just parameter way to handle options.

PR #287, opened 2016-02-19
Added enable_gateway_ports_connections()

rplevka:
Allows the remote connections to SSH remote port forwards.
This will be used mainly for creating fake proxies.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2582,3 +2582,9 @@ def relink_manifest(manifest_file=None):
     run('unlink /opt/manifests/manifest-latest.zip')
     run('ln -s {0} /opt/manifests/manifest-latest.zip'
         .format(new_manifest_file))
+
+
+def enable_gateway_ports_connections():
+    """Required by remote connections to SSH tunnels"""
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -2582,3 +2582,9 @@ def relink_manifest(manifest_file=None):
     run('unlink /opt/manifests/manifest-latest.zip')
     run('ln -s {0} /opt/manifests/manifest-latest.zip'
         .format(new_manifest_file))
+
+
+def enable_gateway_ports_connections():
+    """Required by remote connections to SSH tunnels"""
+    run('sed "s/^[#]*GatewayPorts.*/GatewayPorts yes/" /etc/ssh/sshd_config')
```
elyezer: This will not change the config file unless the `-i` (in place) option is added. I would recommend adding an argument to the option to make sed create a backup file.

elyezer:
ACK, pending comment.

automation_tools/__init__.py
rplevka: :+1: good catch. I usually use it without `-i` first to make sure it works well and add later. I simply forgot that here.

rplevka:
comments addressed (`-i` added to the `sed` command)

abalakh:
ACK

PR #288, opened 2016-02-22
Fix RHEL6 Silent Fail

jyejare:
Fixes the silent Upgrade Job failed during 'YUM UPDATE' on rhel6.


elyezer:
ACK

sthirugn:
default value of `quiet` is `False`, but I am okay with calling it `False` explicitly too to make it more readable.

sthirugn:
ACK

Issue #267, opened 2015-11-25
Task install_puppet_scap_client fails when installing Satellite6.0


rwilliams5:
product_install task fails to install package puppet-foreman_scap_client. Looks like this package does not exist within the repos set for satellite6.0.

Maybe there should be a check around this task depending on the OS version?

omaciel:
@rwilliams5 yeah, iirc there is no `puppet-foreman_scap_client` for RHEL 6 so a check here would make sense. @elyezer ?

elyezer:
I will work on this. Thanks @rwilliams5 for the report.

kbidarkar:
There is no such package for Satellite6.0 , but only from Sat6.1 ( only downstream, upstream needs a tar file I believe) . Need to investigate in detail further. Will do.

rplevka:
https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L1148
this is the place we trigger `oscap client` installation. We do not check which version of satellite 6 is being installed. We need to add a conditional to trigger this only if `6.0` is being installed

PR #293, opened 2016-02-24
Install or configure things related to OSCAP only if it's Sat6.1

kbidarkar:


kbidarkar:
Fixes, https://github.com/SatelliteQE/automation-tools/issues/267

automation_tools/__init__.py
```diff
@@ -1144,8 +1144,9 @@ def product_install(distribution, create_vm=False, certificate_url=None,
                 execute(setup_foreman_discovery, host=host)
             if os.environ.get('LIBVIRT_KEY_URL') is not None:
                 execute(setup_libvirt_key, host=host)
-            execute(install_puppet_scap_client, host=host)
-            execute(setup_oscap, host=host)
+            if os.environ.get('SATELLITE_VERSION') == '6.1':
```
rplevka: perhaps we might change the conditional to include all 6.1+ versions (with 6.2 being released soon) and only exclude 6.0.

rplevka:
ACK pending comment

automation_tools/__init__.py
kbidarkar: Let's do that when we add support for Sat 6.2 and not now. Currently SATELLITE_VERSION just supports version 6.0 and 6.1

kbidarkar: or may be let me change it to `!= '6.0'`

sthirugn:
ACK

Issue #105, opened 2014-12-22
setup_default_capsule should handle situations when a FQDN is not present


omaciel:
If you don't configure a server with a valid ``fqdn``, then ``setup_default_capsule`` will fail with the following stack trace:

```bash
Executing task 'setup_default_capsule'
Traceback (most recent call last):
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/main.py", line 743, in main
    *args, **kwargs
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 384, in execute
    multiprocessing
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 274, in _execute
    return task.run(*args, **kwargs)
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 174, in run
    return self.wrapped(*args, **kwargs)
  File "/home/jenkins/workspace/satellite6-installer/automation_tools/__init__.py", line 968, in product_install
    execute(setup_default_capsule, host=host)
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 384, in execute
    multiprocessing
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 274, in _execute
    return task.run(*args, **kwargs)
  File "/home/jenkins/shiningpanda/jobs/90156090/virtualenvs/d41d8cd9/lib/python2.7/site-packages/fabric/tasks.py", line 174, in run
    return self.wrapped(*args, **kwargs)
  File "/home/jenkins/workspace/satellite6-installer/automation_tools/__init__.py", line 302, in setup_default_capsule
    domain = hostname.split('.', 1)[1]
IndexError: list index out of range
```

elyezer:
@omaciel what you suggest to do when this happens? Abort the installation or try some workaround?

omaciel:
@elyezer maybe default to `localhost.localdomain`?

elyezer:
[...]

PR #294, opened 2016-02-24
Make sure the hostname is configured with a FQDN

kbidarkar:


kbidarkar:
Fixed in https://github.com/SatelliteQE/automation-tools/issues/105

automation_tools/__init__.py
```diff
@@ -244,6 +244,9 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
     if len(hostname) == 0:
         print('Was not possible to fetch hostname information')
         sys.exit(1)
+    if '.' not in hostname:
```
rplevka: perhaps we could merge the checks and have it like:
```python
if '.' not in hostname or len(hostname.split('.', 1)[1]) == 0:
    print('Please make sure the hostname is configured with a valid FQDN')
    sys.exit(1)
```

elyezer: Better to have separated otherwise will be required to repeat `hostname.split('.', 1)[1]`

elyezer:
ACK

sthirugn:
ACK

sthirugn:
I will let @rplevka review again and merge.

rplevka:
@sthirugn, thanks :)
ACK

Issue #296, opened 2016-02-24
make yum calls with lower verbosity level to hide progress bars


rplevka:
`automation_tools` scripts executing yum provide quite many lines of (mostly) useless output. Especially when it comes to download progress bars

```bash
$ yum -y install cas
Loaded plugins: product-id, search-disabled-repos, security, subscription-manager
Setting up Install Process
Resolving Dependencies
--> Running transaction check
---> Package cas.noarch 0:0.15-1.el6.1 will be installed
--> Processing Dependency: python-paramiko for package: cas-0.15-1.el6.1.noarch
--> Running transaction check
---> Package python-paramiko.noarch 0:1.7.5-2.1.el6 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

==============================================================================================================
 Package                      Arch                Version                    Repository                  Size
==============================================================================================================
Installing:
 cas                          noarch              0.15-1.el6.1               myRepo              119 k
Installing for dependencies:
 python-paramiko              noarch              1.7.5-2.1.el6              myRepo              728 k

Transaction Summary
==============================================================================================================
Install       2 Package(s)

Total download size: 846 k
Installed size: 8.0 M
Downloading Packages:
(1/2): cas-0.15-1.el6.1.noarch.rpm                                                     | 119 kB     00:00     
(2/2): python-paramiko-1.7.5-2.1.el6.noarch.rpm                                        | 728 kB     00:00     
--------------------------------------------------------------------------------------------------------------
Total                                                                         3.8 MB/s | 846 kB     00:00     
Running rpm_check_debug
Running Transaction Test
Transaction Test Succeeded
Running Transaction
  Installing : python-paramiko-1.7.5-2.1.el6.noarch                                                       1/2 
  Installing : cas-0.15-1.el6.1.noarch                                                                    2/2 
  Verifying  : python-paramiko-1.7.5-2.1.el6.noarch                                                       1/2 
  Verifying  : cas-0.15-1.el6.1.noarch                                                                    2/2 

Installed:
  cas.noarch 0:0.15-1.el6.1                                                                                   

Dependency Installed:
  python-paramiko.noarch 0:1.7.5-2.1.el6                                                                      

Complete!
```
What's more, fabric output renders every update on a single line (increased percentage, appended '=' sign) as a standalone output line.

We might save ourselves from tons of useless logs by simply specifying a lower debug level for yum command:
```bash
$ yum -y -d1 install cas

====================================================================================================
[...]

PR #297, opened 2016-02-25
set_yum_debug_level() function added

rplevka:
implements https://github.com/SatelliteQE/automation-tools/issues/296
the  function that sets `debuglevel` value in `/etc/yum/yum.conf`.
Defaults to `1` which hides some progress bar graphics and makes yum output less talky.


elyezer:
@rplevka isn't a way to change the verbosity on the command line? If not we can start changing the config as you proposed on this PR.

rplevka:
@elyezer, yes, there is. `-d` parameter. However I was thinking we might set it to `1` globally to have this as a default behaviour for all `yum` calls and the few we might want to be verbose we can set from a command line manually.

elyezer:
@rplevka just wanted to know if there is the way. We can start using it on the product_install. Thanks for the clarification.

elyezer:
ACK

jyejare:
ACK

sthirugn:
:+1:

PR #299, opened 2016-02-26
Added support for running installer for Satellite6.2

kbidarkar:
a) The changes are compatible for Sat6.0 and Sat6.1
   installations.
b) SATELLITE_VERSION env variable is mandatory and needs to be
   set for Satellite6.2 installation and will no longer be used
   just for provisioning.
c) katello-installer or foreman-installer will be invoked
   depending upon the SATELLITE_VERSION ENV variable.
d) Capsule Features will get configured as 'capsule-dns' or
   'foreman-proxy-dns' depending upon the SATELLITE_VERSION
   ENV variable.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -274,30 +274,36 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
             print('Was not possible to fetch interface information')
             sys.exit(1)
 
+    proxy = 'capsule'
+    if os.environ.get('SATELLITE_VERSION') == '6.2':
+        proxy = 'foreman-proxy'
+
     installer_options = {
-        'capsule-dns': 'true',
-        'capsule-dns-forwarders': forwarders,
-        'capsule-dns-interface': interface,
-        'capsule-dns-zone': domain,
-        'capsule-dhcp': 'true',
-        'capsule-dhcp-interface': interface,
-        'capsule-tftp': 'true',
-        'capsule-tftp-servername': hostname,
+        '{0}-dns'.format(proxy): 'true',
+        '{0}-dns-forwarders'.format(proxy): forwarders,
+        '{0}-dns-interface'.format(proxy): interface,
+        '{0}-dns-zone.format(proxy)': domain,
+        '{0}-dhcp'.format(proxy): 'true',
+        '{0}-dhcp-interface'.format(proxy): interface,
+        '{0}-tftp'.format(proxy): 'true',
+        '{0}-tftp-servername'.format(proxy): hostname,
         'capsule-puppet': 'true',
-        'capsule-puppetca': 'true',
-        'capsule-register-in-foreman': 'true',
+        '{0}-puppetca'.format(proxy): 'true',
+        '{0}-register-in-foreman'.format(proxy): 'true',
     }
 
     if 'DHCP_RANGE' in os.environ:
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2600,4 +2614,6 @@ def enable_gateway_ports_connections():
 
 def set_yum_debug_level(level=1):
     """Set default debug level for yum output"""
-    run('sed -i "s/^[#]*debuglevel=.*/debuglevel={}/" /etc/yum.conf'.format(level))
+    run('sed -i "s/^[#]*debuglevel=.*/debuglevel={}/" /etc/yum.conf'.format(
+        level
```
kbidarkar: changes for flake8

kbidarkar:
Need to raise robottelo-ci side PR and update the SATELLITE_VERSION part.

automation_tools/__init__.py
elyezer:
```python
run(
    'sed -i "s/^[#]*debuglevel=.*/debuglevel={}/" /etc/yum.conf'
    .format(level)
)
```

?

automation_tools/__init__.py
```diff
@@ -1722,19 +1730,25 @@ def katello_installer(debug=False, sam=False, verbose=True, **kwargs):
...
+        forwarders = kwargs.pop('{0}-dns-forwarders'.format(proxy))
         for forwarder in forwarders:
             extra_options.append(
-                '--capsule-dns-forwarders="{0}"'.format(forwarder))
+                '--{0}-dns-forwarders="{1}"'.format(proxy, forwarder))
 
-    run('{0}-installer {1} {2} {3} {4}'.format(
-        'sam' if sam else 'katello',
+    run('{0}-installer {1} {2} {3} {4} {5}'.format(
+        'sam' if sam else installer,
```
elyezer: Better to separate the installer selection in a if/elif/else block.

automation_tools/__init__.py
```diff
@@ -1722,19 +1730,25 @@ def katello_installer(debug=False, sam=False, verbose=True, **kwargs):
...
     """
-    # capsule-dns-forwarders should be repeated if setting more than one value
-    # check if a list is being received and repeat the option with different
-    # values
+    # capsule-dns-forwarders should be repeated if setting more than one
+    # value check if a list is being received and repeat the option with
+    # different values
+    proxy = 'capsule'
+    installer = 'katello'
+    if os.environ.get('SATELLITE_VERSION') == '6.2':
```
elyezer: We should leave this helper function as a helper function, that said environment variables should be handled outside. Please add a parameter called satellite_version (default to `'6.2'`?) and handle the environment variable on product_install, also product_install should fail early if that variable is not defined.

elyezer:
ACK, pending comments.

automation_tools/__init__.py
kbidarkar: ok

automation_tools/__init__.py
kbidarkar: ok, will update

automation_tools/__init__.py
```diff
@@ -865,6 +871,8 @@ def downstream_install(admin_password=None, run_katello_installer=True):
 
     # Install required packages for the installation
     run('yum install -y katello')
```
jyejare: This will run in case of version 6.2 also, please put this in else clause below.

jyejare:
ACK.

Issue #298, opened 2016-02-26
Allow RHEL Vault automation runs


lpramuk:
As Satellite is layered product we need from time to time ensure its compatibility with new RHEL dot release candidates. By enable such candidate repo before OS upgrade and SAT installation we could achieve this.

rplevka:
APT

sthirugn:
ACK pending PR :)

lpramuk:
when we have started acking issues, huh? :)

sthirugn:
ha ha I just followed @rplevka ;)

PR #300, opened 2016-02-26
enable beaker repo for OS upgrade before sat6 install

lpramuk:
As Satellite is layered product we need from time to time ensure its compatibility with new RHEL dot release candidates. Having possibility to enable such candidate repo before OS upgrade and SAT installation we could achieve it. 

Implements #298

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -20,7 +20,9 @@ from urlparse import urlsplit
 
 from automation_tools.satellite6.capsule import generate_capsule_certs
 from automation_tools.repository import (
-    delete_custom_repos, enable_satellite_repos, enable_repos, disable_repos)
+    create_custom_repos, delete_custom_repos, enable_satellite_repos,
+    enable_repos, disable_repos
+)
 from automation_tools.utils import distro_info, update_packages
 from fabric.api import cd, env, execute, get, local, put, run, settings, sudo
 from novaclient.client import Client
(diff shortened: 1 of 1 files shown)


lpramuk:
I need to check what happens if I leave OS_UPGRADE_REPO field empty....

automation_tools/__init__.py
```diff
@@ -1074,7 +1076,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         cdn_version=sat_cdn_version,
         host=host
     )
-
+    # Create custom beaker repo
+    if os.environ.get('OS_UPGRADE_REPO'):
+        beaker_repo_url = os.environ.get('OS_UPGRADE_REPO')
```
elyezer: s/beaker_repo_url/os_upgrade_repo_url

automation_tools/__init__.py
```diff
@@ -1074,7 +1076,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+        execute(create_custom_repos, beaker_server=beaker_repo_url, host=host)
```
elyezer: There is no `beaker_server` argument being handled on create_custom_repos.

elyezer:
ACK, pending comments

automation_tools/__init__.py
elyezer: Maybe we should also rename the envvar to `OS_UPGRADE_REPO_URL`

elyezer: Also document the new env var on the docstring.

lpramuk: np

automation_tools/__init__.py
lpramuk: True, but there is **kwargs argument. So it can be any argument.
" ... Where the key is the repository filename and repository name, and the value is the
    repository URL."

elyezer: @lpramuk never mind, I was a long time away of this code that I forgot that it is going to create the repositories based on the kwargs.

elyezer:
ACK, pending document OS_UPGRADE_REPO on the product_install docstring.

lpramuk:
Comments resolved.

elyezer:
ACK

sthirugn:
ACK

PR #307, opened 2016-03-14
local env var overrides fabric env issue

jyejare:
Local 'env' variable replacing / overriding fabric 'env'.
Renamed local 'env' to 'envr'

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1931,10 +1931,11 @@ def sync_capsule_tools_repos_to_upgrade(admin_password=None):
     version = distro_info()[1]
     details = os.environ.get('CAP_SUB_DETAILS')
     if details is not None:
-        cv_name, env, ak_name = [item.strip() for item in details.split(',')]
+        cv_name, env_name, ak_name = [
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1931,10 +1931,10 @@ def sync_capsule_tools_repos_to_upgrade(admin_password=None):
     version = distro_info()[1]
     details = os.environ.get('CAP_SUB_DETAILS')
     if details is not None:
-        cv_name, env, ak_name = [item.strip() for item in details.split(',')]
+        cv_name, envr, ak_name = [item.strip() for item in details.split(',')]
```
abalakh: why not `env_name` to consist with 2 others? or it's used too?
and btw, in `envr`, what does `r` stand for?

abalakh:
ACK

automation_tools/__init__.py
jyejare: Sounds good 'env_name', I was not getting a good name and so 'envr' :)

sthirugn:
ACK

PR #308, opened 2016-03-16
Capsule provisioning

elyezer:
Add setup_capsule_firewall task and update the setup_firewall to
setup_satellite_firewall. The setup_firewall task now is a helper for
setting up firewall rules and using iptables on RHEL6 and firewall-cmd
on RHEL7.

Add setup_capsule_content task to help setting up a custom content on
the Satellite Server that allow creating a capsule.

Update capsule_installer to match the new way to run the capsule
installer.

Update setup_capsules to not start content sync. This may be added later
if syncing is something really needed.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2,7 +2,6 @@
 
 Many commands are affected by environment variables. Unless stated otherwise,
 all environment variables are required.
-
 """
 from __future__ import print_function
 import os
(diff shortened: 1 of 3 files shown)


automation_tools/satellite6/capsule.py
```diff
@@ -393,7 +611,7 @@ def setup_capsules(path):
         with settings(host_string=capsule.host_string):
             register_capsule()
             put(local_path=cert_path)
-            capsule_installer(*oauth_info)
+            capsule_installer(capsule.hostname, cert_path, *oauth_info)
 
     # Start content synchronization in all registered capsules
```
kbidarkar: We can now do syncing from the WEBUI too, so I feel we should have a arg/flag to enable ot disable it.

sthirugn:
It is sad to see that hammer commands need to be rewritten here.  I understand that the intent here is not to make automation-tools dependent on nailgun or other tools.  But since we are using hammer, we may need to take care of its changes according to satellite versions, if they happen to differ in future due to some reason.  Is there an other easier alternative - like using `nailgun`?  perhaps keeping it as an optional dependency and use it only when installing capsule? just thinking out loud...

automation_tools/satellite6/capsule.py
```diff
@@ -225,7 +217,8 @@ def register_capsule():
     """Register the capsule on the Satellite 6 server."""
     config = _get_config()
     run(
-        'rpm -Uvh http://{}/pub/katello-ca-consumer-latest.noarch.rpm'
+        'yum -y localinstall '
+        'http://{}/pub/katello-ca-consumer-latest.noarch.rpm'
```
sthirugn: just curious, what is the need for this change?

sthirugn:
ACK pending question.  Note: my question is not a blocker.

automation_tools/satellite6/capsule.py
elyezer: It is needed since this is the way the UI page recommends customers to install the certificates RPM.

sthirugn: Which UI page you mean? I see latest Hosts -> Content Hosts page still refers to rpm command.

elyezer: @sthirugn sorry, it is from the output from the command you use to generate the certificates `capsule-certs-generate`.

sthirugn: @elyezer no worries, ty for the explanation :+1:

elyezer:
@kbidarkar @sthirugn if you are ok with this PR, can you please merge it?

elyezer:
Rebased

sthirugn:
This already has my ACK.  Can somebody else do a second ack/merge please?

automation_tools/__init__.py
```diff
@@ -353,63 +352,124 @@ def setup_fake_manifest_certificate(certificate_url=None):
...
+                'udp': (
+                    4242,
+                    4422,
+                ),
+            }
+    """
+    if definitions is None:
+        return
+
+    if distro_info()[1] >= 7:
```
omaciel: Should this be `6` instead of `7`?

automation_tools/satellite6/capsule.py
omaciel: I thought `localinstall` was needed only if the package was local to the filesystem where the command is being run?

omaciel:
@elyezer couple of comments

automation_tools/__init__.py
elyezer: Good catch, this actually needs to be < 7 or == 6

elyezer: Like here https://github.com/SatelliteQE/automation-tools/pull/308/files#diff-a692ae115cc45cc7627d4545e9b4c294R394

automation_tools/satellite6/capsule.py
elyezer: localinstall have the same effect as rpm -Uvh. When  passing an URL it's gonna first download the RPM and then install it locally

elyezer:
Comments addressed. Thanks @omaciel for the eagle eyes.

omaciel:
ACK

PR #312, opened 2016-03-18
Fix #310 calling disable_repos for 'extras' after Docker installation

### LINKED ISSUE #310: Disable 'extras' repository after installing Docker on RHEL 7 systems
We should add another command after [this](https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L195) to make sure that we disable the 'extras' repository after installing Docker on RHEL 7 systems. Note that we do that already for RHEL 6 systems.

The new line should be:

```python
        run(
            'subscription-manager repos --disable "rhel-{0}-server-extras-rpms"'
            .format(os_version)
        )
```

rochacbruno:
Implemented using **disable_repos** function

tested with @elyezer 's help

output

```
[root@...] run: subscription-manager repos --disable "rhel-7-server-extras-rpms"
[root@...] out: Repository 'rhel-7-server-extras-rpms' is disabled for this system. 
```

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -194,6 +194,8 @@ def setup_default_docker():
     # Install ``Docker`` package.
     if os_version >= 7:
         run('yum install -y docker', warn_only=True)
+        # Disable 'extras' repo after installing Docker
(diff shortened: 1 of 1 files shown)


elyezer:
ACK

PR #317, opened 2016-04-04
Added set_yum_debug_level call to satellite6_upgrade and product_install

rplevka:
This should save us from a long useless yum stdout. It will set yum loglevel to 1 which only prints out a list summary of packages that have been manipulated (no progressbars, etc.). Yum also runs faster

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1166,6 +1166,9 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         execute(update_rhsm_stage, host=host)
     execute(subscribe, host=host)
 
+    # Setting yum stdout log level to be less verbose
+    execute(set_yum_debug_level)
+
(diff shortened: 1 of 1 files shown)


jyejare:
ACK

lpramuk:
ACK

elyezer:
@jyejare @lpramuk @rplevka this PR is not good... need to pass `host=host`. I'm gonna fix this, we need to pay attention since a change on product_install can break a lot of jobs.

elyezer:
Added 60a39113724d33b61a090a00736a2123b3540caf to fix/update it.

rplevka:
@elyezer :100:  can't understand how on earth i could miss that..thanks!

elyezer:
@rplevka np.

vpilat:
OT: mas u me postu... ;)

rplevka:
@vpilat: ?

PR #319, opened 2016-04-18
Upgrade Automation on RHEV

jyejare:
As planned earlier and due to non availability of provisioning on Openstack project we decided to move on rhevm. And so the reason we need to have functions for creating and deleting instances created from template on rhevm. This PR includes these two functions.

Note: Here in RHEVM template is same as openstack image.

While saving a template of any Sat server, one should save network as well as storage configuration with template otherwise this functions will fail to do an upgrade job for us(as of now, we ll add these configurations by ourself later or may keep as it is for simplicity). Also one should add jenkins ssh key in template before saving and using that template to create instance from it.
I am still in research on how to add given shh key by automation.

So lets review the PR and Welcome RHEVM in Satellite Upgrade Automation.

Suggestions for improvements are always welcome.


automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+    The following environment variables affect this command:
+
+    USERNAME
+        The username of a rhevm project to login.
+    PASSWORD
+        The password of a rhevm project to login.
+    API_URL
+        An url to API of rhevm project.
+    """
+    username = os.environ.get('USERNAME')
```
elyezer: USERNAME and PASSWORD is too generic names, we should use a more specific ones.

automation_tools/__init__.py
```diff
@@ -25,6 +25,9 @@
 from automation_tools.utils import distro_info, update_packages
 from fabric.api import cd, env, execute, get, local, put, run, settings, sudo
 from novaclient.client import Client
+from ovirtsdk.api import API
```
elyezer: ovirtsdk is a new import and will not be installed.

We need add this as an requirement.

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
     return output[ip]
 
 
+def get_rhevm_client():
```
elyezer: The __init__.py file is to big already, what about start a new module for these tasks/functions? Eventually we also would move all the upgrade related tasks to that module.

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+    if username is None:
+        print('The USERNAME environment variable should be defined.')
+    password = os.environ.get('PASSWORD')
+    if password is None:
+        print('The PASSWORD environment variable should be defined.')
+    api_url = os.environ.get('API_URL')
+    if api_url is None:
+        print('An API_URL environment variable should be defined.')
+    try:
+        obj = API(
```
elyezer: `return API(...`

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+            url=api_url,
+            username=username,
+            password=password,
+            insecure=True
+        )
+        return obj
```
elyezer: Drop this line when implementing the comment above.

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+
+    USERNAME
+        The username of a rhevm project to login.
+    PASSWORD
+        The password of a rhevm project to login.
+    API_URL
+        An url to API of rhevm project.
+
+    """
+    obj = get_rhevm_client()
```
elyezer: s/obj/rhevm_client?

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
     return output[ip]
 
 
+def get_rhevm_client():
+    """Creates client object instance for rhevm.
```
elyezer: Creates and return a client for RHEVM

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+
+    Returns the client object for further use.
```
elyezer: Remove this information and state that the caller of this function should take care of closing the client.

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+        The username of a rhevm project to login.
+    PASSWORD
+        The password of a rhevm project to login.
+    API_URL
+        An url to API of rhevm project.
+
+    """
+    obj = get_rhevm_client()
+    temp = obj.templates.get(name=template_name)
+    dc = obj.datacenters.get(name='Default')
```
elyezer: What about adding an argument defaulting to `Default` for the datacenters?

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+    quota = dc.quotas.get(name='admin')
```
elyezer: What about an argument default to `admin` for this information?

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+            name=instance_name,
+            cluster=obj.clusters.get('Default'),
+            template=temp, quota=quota))
+    timeup = time.time() + int(timeout) * 60
+    while True:
+        if time.time() > timeup:
+            print('The timeout for creating RHEVM instance has reached!')
+            sys.exit(1)
+        vm_status = obj.vms.get(name=instance_name).get_status().get_state()
+        print(vm_status)
```
elyezer: Create a more informational string like "Waiting for RHEVM instance, current status {0}"

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+    quota = dc.quotas.get(name='admin')
+    print('Turning on instance {0} from template {1}. Please wait'
+          'till it get up ...'.format(instance_name, template_name))
+    obj.vms.add(
+        params.VM(
+            name=instance_name,
+            cluster=obj.clusters.get('Default'),
+            template=temp, quota=quota))
+    timeup = time.time() + int(timeout) * 60
+    while True:
```
elyezer: This should be refactored the chained while True could be very dangerous and lead to deadlock.

```python
while True:
    if time.time() > timeup:
        print('The timeout for creating RHEVM instance has reached!')
        sys.exit(1)
    vm_status = obj.vms.get(name=instance_name).get_status().get_state()
    if vm_status == 'down':
        obj.vms.get(name=instance_name).start()
        continue
    if vm_status == 'up':
        break
    time.sleep(3)
```

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+                    sys.exit(1)
+                vm_status = obj.vms.get(
+                    name=instance_name).get_status().get_state()
+                print(vm_status)
+                if vm_status == 'up':
+                    break
+                time.sleep(5)
+            break
+        time.sleep(3)
+    print('Instance {0} is now UP !'.format(instance_name))
```
elyezer: s/UP !/up!

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+    obj.disconnect()
+
+
+def delete_rhevm_instance(instance_name, timeout=5):
+    """Deletes RHEVM Instance.
+
+    :param instance_name: A string. RHEVM instance name to delete.
```
elyezer: Move parameters description to the end of the docstring.

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+        sys.exit(1)
+
+
+def create_rhevm_instance(instance_name, template_name, timeout=5):
+    """Creates rhevm Instance from template.
+
+    The assigning template should have network and storage configuration saved
+    already.
+
+    :param instance_name: A string. RHEVM Instance name to create.
```
elyezer: Move the parameter description to the end of the docstring.

automation_tools/__init__.py
```diff
@@ -2127,6 +2130,151 @@ def get_hostname_from_ip(ip, timeout=3):
...
+    """
+    obj = get_rhevm_client()
+    vm = obj.vms.list(query='name={0}'.format(instance_name))
+    if not vm:
+        print('The instance {0} is not found '
+              'in RHEVM to delete!'.format(instance_name))
+    else:
+        print('Deleting instance {0} from RHEVM.....'.format(instance_name))
+        timeup = time.time() + int(timeout) * 60
+        if obj.vms.get(name=instance_name).get_status().get_state() == 'up':
```
elyezer: We should try to make this code simpler. I'm surprised that the client don't handles the deletion and a pooling is needed.

I will try to think on a better way to handle this, but at first I think the waiting for something can be abstracted into a function for waiting for something to happen.

automation_tools/__init__.py
jyejare: As we are going to use the defined datacenter and quata for upgrade on rhevm, so I didnt make it optional.
But still It will be good for other feature testing in future.

automation_tools/__init__.py
jyejare: So the reason I have timeout so that It can max try for 5 minutes.

jyejare:
@elyezer , Sorry I was very excited to add this in our upgrade automation so just posted my rough code as it is and so the reason I made a mistakes with some variable names and etc.

As of now I have just pull out code from automation_tools/__init__.py and put into automation_tools/satellite6/upgrade.py without creating any classes. I will handle it neatly later or if possible then in this PR only. Please let me know your further comments.

elyezer:
ACK

PR #322, opened 2016-04-21
Updated discovery setup to install required packages

sghai:
The required discovery plugin doesn't get installed automatically, we need to install it manually to enable UI menu options for 'Discovered Hosts' and discovery rules.

Also mentioned packages installs discovery image and enables the feature on smart-proxy.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -558,6 +558,18 @@ def install_puppet_scap_client():
 
 def setup_foreman_discovery():
     """Task to setup foreman discovery."""
+    # Install required packages to enable discovery plugin
+    packages = (
+        'tfm-rubygem-foreman_discovery',
+        'rubygem-smart_proxy_discovery',
+        'foreman-discovery-image',
+        'tfm-rubygem-hammer_cli_foreman_discovery',
+    )
+    run('yum install -y {0}'.format(' '.join(packages)))
+
+    for daemon in ('foreman', 'httpd', 'foreman-proxy'):
+        manage_daemon('restart', daemon)
+
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
     template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
     if template_url is None:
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -558,6 +558,19 @@ def install_puppet_scap_client():
 
 def setup_foreman_discovery():
     """Task to setup foreman discovery."""
+    # Install required packages to enable discovery plugin
+    packages = [
+        'tfm-rubygem-foreman_discovery'
```
omaciel: separate package names with a comma (ie. `'tfm-rubygem-foreman_discovery',`)

automation_tools/__init__.py
```diff
@@ -558,6 +558,19 @@ def install_puppet_scap_client():
...
+        'rubygem-smart_proxy_discovery'
+        'foreman-discovery-image'
+        'tfm-rubygem-hammer_cli_foreman_discovery'
+    ]
+    for package in packages:
```
omaciel: By the way, you could get them all installed in one shot (and not use a loop) by doing something like:

```python
run('yum install -y {0}'.format(" ".join(package)))
```

omaciel:
ACK pending comment (by the way, don't block on me to merge this. I was lucky to find 5m to look at this PR today :) )

sghai:
Thanks @omaciel Comments have been addressed.

sthirugn:
wait, are we not supposed to use these installer options which should do all these for us?

# foreman-installer --help | grep discovery
    --[no-]enable-foreman-plugin-discovery Enable 'foreman_plugin_discovery' puppet module (default: false)
= Module foreman_plugin_discovery:
    --foreman-plugin-discovery-image-name  tarball with images (default: "fdi-image-latest.tar")
    --foreman-plugin-discovery-install-images  should the installer download and setup discovery images
    --foreman-plugin-discovery-source-url  source URL to download from (default: "http://downloads.theforeman.org/discovery/releases/latest/")
    --foreman-plugin-discovery-tftp-root  tftp root to install image into (default: "/var/lib/tftpboot")

sghai:
Yes, we can do this.. but here `--enable-foreman-plugin-discovery` only installs one package `tfm-rubygem-foreman_discovery`. we need other packages like one for image other for smart_proxy. 

Other options to install image and set tftp-root doesn't work as expected. I'll file a bz for this.

sthirugn:
ACK

PR #326, opened 2016-04-21
Real fix for Bug 1329394 - Skip oscap install for rhel 6

sthirugn:


sthirugn:
There is no other QE available online at this time :/

I am merging this to retrigger rhel 6 automation.

omaciel:
@sthirugn why we're not checking for Satellite 6.2 anymore and only for 6.1? So we always want to run `install_puppet_scap_client` but only run `setup_oscap` fro RHEL 7?

elyezer:
Running `distro_info` that way is actually running on the machine where is provisioning the satellite server, not on the server at all. I'm working on a fix.

elyezer:
Sent PR #328 to fix this.

sthirugn:
@omaciel: Yes. The summary is: 
* for 6.0 we don't even run setup_oscap because it is not supported
* for 6.1, we install oscap packages separately
* for 6.2, we use the installer to install
* due to open bug, temporarily we skip installing oscap in rhel 6
ping me if you have more questions :)

omaciel:
@sthirugn so we only want to run `install_puppet_scap_client` for 6.1 and not for 6.0 (unsupported) or 6.2 (use the installer instead)? Seems that the conditional statement will run `install_puppet_scap_client` for 6.2 as well, which confuses me a bit :/

sthirugn:
you are right and this is how it works now.  afaik @kbidarkar is yet to do some refactoring around install_puppet_scap_client()

PR #327, opened 2016-04-22
Enable SSH GatewayPorts on product_install

elyezer:
This is required in order to create reverse SSH tunnels that are accessible
from others machines not just the local machine.


sthirugn:
ACK and merge

rplevka:
@elyezer :+1: 
..i was just about to submit that 0:)

elyezer:
@rplevka since I did not get your position about it and the new build is about to be out I decided to send this. I hope this will bring better results.

PR #332, opened 2016-05-05
Added support for beta vs ga releases

sthirugn:
Testing is blocked and is reported

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -955,6 +955,8 @@ def downstream_install(admin_password=None, run_katello_installer=True):
         Optional, defaults to 'changeme'. Foreman admin password.
     BASE_URL
         URL for the compose repository.
+    SATELLITE_VERSION
+        Satellite version.
 
     """
     if admin_password is None:
@@ -1120,6 +1122,13 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     :param str certificate_url: where to fetch a fake certificate.
 
     """
+    # Fetch the Satellite Release information.
+    satellite_release = os.environ.get('SATELLITE_RELEASE', 'GA')
+    # SATELLITE_RELEASE should only be BETA or GA
+    if satellite_release.lower() not in ('beta', 'ga'):
+        print('Invalid release provided.  Values allowed: BETA, GA')
+        sys.exit(1)
+
     # Command-line arguments are passed in as strings.
     if isinstance(create_vm, str):
         create_vm = (create_vm.lower() == 'true')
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1178,6 +1180,9 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     # Fetch the Satellite Version information.
     satellite_version = os.environ.get('SATELLITE_VERSION')
 
+    # Fetch the Satellite Release information.
+    satellite_release = os.environ.get('SATELLITE_RELEASE')
```
elyezer: if the release is not available what should we do?

automation_tools/__init__.py
```diff
@@ -1261,6 +1266,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         host=host,
         sam=distribution.startswith('sam'),
         sat_version=satellite_version,
+        sat_release=satellite_release,
```
elyezer: Is this prepared to receive `None` as release? Otherwise we should make sure a proper release is passed.

elyezer: I commenting before reading the remaining of the code. Never mind about this.

elyezer:
I think the solution for this would be simpler. Since the package installed is the same what about checking if the satellite-installer or katellite-installer or foreman-installer is available? This can be handled on the installer task without the need of adding more required arguments and environment variables.

What do you think about this?

sthirugn:
I spoke to @elyezer about his recommendation of checking the presence of satellite-installer/katello-installer.  But at this time, it is good to have explicit checks since there are lot of changes happening between beta and GA.  We can merge this PR and plan any improvements for later.  @elyezer also agreed to this in irc.

sthirugn:
ready for final review/merge

automation_tools/__init__.py
```diff
@@ -1863,7 +1869,16 @@ def katello_installer(debug=False, sam=False, verbose=True, sat_version='6.2',
         sat_version = ''
     if sat_version == '6.2':
         proxy = 'foreman-proxy'
-        installer = 'foreman'
+        if sat_release == 'GA':
```
elyezer: what about ignoring the case here? `sat_release.lower() == 'ga'`

elyezer:
ACK pending comment.

automation_tools/__init__.py
```diff
@@ -955,6 +955,8 @@ def downstream_install(admin_password=None, run_katello_installer=True):
         Optional, defaults to 'changeme'. Foreman admin password.
     BASE_URL
         URL for the compose repository.
+    SATELLITE_VERSION
+        Satellite version.
```
sthirugn: This is just docstring update - this is not introduced by this PR.

sthirugn:
Comments updated and rebased

elyezer:
ACK

PR #333, opened 2016-05-06
Reorganize tasks on product_install

elyezer:
Mitigate the current failure of hammer template not being available by
reorganizing the tasks that product_install executes.


omaciel:
ACK

elyezer:
Merged, let's do another automation run.

PR #334, opened 2016-05-06
GA Snap10.0 - temporary workaround for GemConflict

lpramuk:
GA Snap10.0 - temporary workaround for GemConflict

- FIXME: dont forget to remove once fixed

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1263,7 +1263,11 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         sat_version=satellite_version,
         **installer_options
     )
-
+    # FIXME: temporary workaround for GemConflict in Snap10.0
+    execute(lambda: run(
+        "sed -i 's/0\.0\.16/0.0.14/' /opt/theforeman/tfm/root/usr/share/gems"
+        "/specifications/hammer_cli_foreman-0.5.1.6.gemspec"), host=host
+    )
     certificate_url = certificate_url or os.environ.get(
         'FAKE_MANIFEST_CERT_URL')
     if certificate_url is not None:


lpramuk:
```
...
[root@<fqdn>] run: sed -i 's/0\.0\.16/0.0.14/' /opt/theforeman/tfm/root/usr/share/gems/specifications/hammer_cli_foreman-0.5.1.6.gemspec
...
:: [   PASS   ] :: Command 'fab -H root@<fqdn> product_install:satellite6-DOWNSTREAM,sat_cdn_version=6.2,test_in_stage=false' (Expected 0, got 0)
```

omaciel:
Thank you but NACK. We need a real fix from DEV on this one.

lpramuk:
What you gonna do if a fix will be pending for a week or so?
This is for unblocking us very very quickly

lpramuk:
Hey, this is just temporary... I wont forget about it I swear :-)

Anyway,  I plan to have a proper place in our automation for such type of fixes.

rplevka:
@omaciel , Perhaps we could trigger the automation against @lpramuk branch so we can still benefit from the fix and don't need to push it to master.
However I think our trigger jobs don't support that at the moment.

omaciel:
At this point, if we're blocked because of a regression after a BETA release, we need a real fix from DEV. If it takes them 1 week to deliver it, we slip the GA by 1 week.

elyezer:
We should definitely be more strict until GA.

elyezer:
@lpramuk it seems that this is not needed anymore. Let's close if that is the case.

elyezer:
Closing this since it seems not valid anymore. If needed it can be reopened.

PR #338, opened 2016-05-21
revert a setting of env PYCURL_SSL_LIBRARY in setup.py

lpramuk:
revert a setting of env PYCURL_SSL_LIBRARY in setup.py

- PYCURL_SSL_LIBRARY needs to be set in parent environment to actually work
  (in robottelo-ci)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import os
-import subprocess
-import sys
 
 try:
     from setuptools import setup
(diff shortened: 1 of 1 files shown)


elyezer:
Since you have not assigned the task I went ahead and created #339. I think we should close this in favor of the other because it also updates the documentation to avoid surprises to the one installing automation-tools.

Other way is to merge this, then I rebase my PR to just update the docs.

lpramuk:
Please merge this and then rebase yours. 
I spent some time on this...

PR #342, opened 2016-05-26
Get the docker clean up script ready to trigger it at the end of automation runs

sthirugn:
Currently:
- the script runs every sunday at 6pm ET and it cleans up all the
containers.  This might potentially affect some running jobs too since the
automation takes a long time to run recently
- If there are any issues in containers in between the week, number of
containers start piling up.

Purpose of this fix:
- this script queries for the containers that are older than 2 days and
stop/remove them.  By this way, the recently created/needed containers wont be
deleted.
- Also after this PR, a new jenkins job will be created and it will be triggered
at the end of every automation run.


elyezer:
ACK

kbidarkar:
ACK

PR #344, opened 2016-05-27
Made changes to allow users to run upgrade without passing image names

sghai:
Currently, if user wants to run upgrade via defined jenkins job then user has to pass Instance hostname and Image name depending on the selected OS. I wanted to make it optional. As if  user doesn't know about the rhev image/hostname, still they can perform upgrade.

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -58,7 +58,7 @@ def satellite6_upgrade(admin_password=None):
     print('Wait till Packages update ... ')
     update_packages(quiet=True)
     # Rebooting the system to see possible errors
-    if os.environ.get('RHEV_SATELLITE'):
+    if os.environ.get('RHEV_SATELLITE') or os.environ.get('SAT_HOST'):
         reboot(120)
     # Setting Satellite61 Repos
     major_ver = distro_info()[1]
@@ -178,7 +178,7 @@ def satellite6_capsule_upgrade(admin_password=None):
     execute(lambda: run("scp -o 'StrictHostKeyChecking no' {0}-certs.tar "
                         "root@{0}:/home/".format(cap_host)), host=sat_host)
     # Rebooting the system again to see possible errors
-    if os.environ.get('RHEV_CAPSULE'):
+    if os.environ.get('RHEV_CAPSULE') or os.environ.get('CAP_HOST'):
         reboot(120)
     # Stopping the services again which started in reboot
     run('for i in qpidd pulp_workers pulp_celerybeat '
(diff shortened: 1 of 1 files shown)


automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -263,9 +263,10 @@ def product_upgrade(
         sys.exit(1)
     if not os.environ.get('SATELLITE_HOSTNAME'):
         if not sat_image and not os.environ.get('RHEV_SATELLITE'):
-            print('Please provide either Satellite RHEVM template name or '
-                  'Satellite HostName to perform upgrade!')
-            sys.exit(1)
+            sat_image = os.environ.get('SAT_IMAGE')
+            sat_host = os.environ.get('SAT_HOST')
```
sghai: these variables `SAT_IMAGE` and `SAT_HOST` are defined in jenkins under `RHEV_Config`

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -293,9 +293,10 @@ def product_upgrade(
     if product == 'capsule':
         if not os.environ.get('CAPSULE_HOSTNAME'):
             if not cap_image and not os.environ.get('RHEV_CAPSULE'):
-                print('Please provide either Capsule RHEVM template name or '
-                      'Capsule HostName to perform upgrade!')
-                sys.exit(1)
+                cap_image = os.environ.get('CAP_IMAGE')
+                cap_host = os.environ.get('CAP_HOST')
```
sghai: these variables `CAP_IMAGE` and `CAP_HOST` are defined in jenkins under `RHEV_Config`

automation_tools/satellite6/upgrade/__init__.py
elyezer: Should be checked after defining sat_image and sat_host if one of them is None and exit. That will help on setting up a new environment and will prevent further failures.

automation_tools/satellite6/upgrade/__init__.py
elyezer: Same here, ideally all required variables and permutations should be checked at the beginning of the task in order to fail early and not run a single command on the target. This also can avoid checking every time a variable is defined since it was already done in the beginning.

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -293,9 +293,10 @@ def product_upgrade(
...
+            else:
+                cap_host = os.environ.get('RHEV_CAPSULE')
```
elyezer: What if `cap_host` receives None (in case RHEV_CAPSULE is not defined)? The task should exit and not proceed on that case.

elyezer:
ACK pending comments.

They are not blocker but will make a huge difference on helping debugging if for some reason a env variable is not defined anymore.

automation_tools/satellite6/upgrade/__init__.py
sghai: :+1:

automation_tools/satellite6/upgrade/__init__.py
sghai: Yeah, In case `RHEV_CAPSULE` is not defined then only we are setting capsule host using:
`cap_host = os.environ.get('CAP_HOST')` 

but if  `RHEV_CAPSULE` is defined then`cap_host = os.environ.get('RHEV_CAPSULE')`

automation_tools/satellite6/upgrade/__init__.py
sghai: +1

sghai:
@elyezer : now checking all env at the beginning of `product_upgrade` task

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -261,11 +261,18 @@ def product_upgrade(
     if product not in products:
         print('Product name should be one of {0}'.format(', '.join(products)))
         sys.exit(1)
+
+    for env_var in ('SAT_IMAGE', 'SAT_HOST', 'CAP_IMAGE', 'CAP_HOST'):
```
elyezer: I would suggest to do something like the following:

```python
missing_vars = [
    var for var in ('SAT_IMAGE', 'SAT_HOST', 'CAP_IMAGE', 'CAP_HOST')
    if var not in os.environ
]
if missing_vars:
    print('The following environment variable(s) must be set: {0}.'.format(', '.join(missing_vars)))
    sys.exit(1)
```

With the above you will know all the missing variables at once and will help setting the missing ones.

sghai: :+1:

elyezer:
ACK

PR #345, opened 2016-05-27
Store docker cleanup log to current folder

sthirugn:
Since the cleanup is triggered from jenkins now, it is better to store the logs in the jenkins workspace instead of giving jenkins user access to system logs.


kbidarkar:
ACK

elyezer:
ACK

PR #348, opened 2016-06-01
updated capsule steps for 6.2 upgrade

sghai:
Added a new step to enable SCL repo for capsule upgrade on rhel6 and that stop services step is valid for 6.0.


elyezer:
ACK

PR #349, opened 2016-06-01
satellite-installer should install oscap by default

kbidarkar:
Satellite-installer should install and configure OSCAP by default,
without having to manually enable it via the isntaller and it's
options.


kbidarkar:
Also updated the setup_oscap function and removed unwanted code,

sthirugn:
ACK

elyezer:
ACK

PR #351, opened 2016-06-06
'katello-service stop' is not required for 6.2 as installer itself stop services

sghai:
In 6.2, installer itself stop services before performing upgrade, so its not required to run `katello-service stop` explicitly.  It should be used only for `6.1`

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -95,9 +95,11 @@ def satellite6_upgrade(admin_password=None):
     # Rebooting the system again for possible errors
     if os.environ.get('RHEV_SATELLITE') or os.environ.get('SAT_HOST'):
         reboot(120)
-    # Stop the service again which started in restart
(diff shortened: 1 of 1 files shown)


jyejare:
ACK

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -95,9 +95,9 @@ def satellite6_upgrade(admin_password=None):
     # Rebooting the system again for possible errors
     if os.environ.get('RHEV_SATELLITE') or os.environ.get('SAT_HOST'):
         reboot(120)
-    # Stop the service again which started in restart
-    run('katello-service stop')
     if to_version == '6.1':
+        # Stop the service again which started in restart
```
sthirugn: can you please add a comment to say that these are already done by satellite-installer in 6.2?

sthirugn:
ACK pending non blocker comment

sghai:
@sthirugn : comment added. thanks for review.

PR #353, opened 2016-06-07
Populate default oscap-content on the satellite6

kbidarkar:
a) This populates rhel6, rhel7 default oscap content for satellite.
b) This is only required for Satellite6.2


automation_tools/__init__.py
```diff
@@ -556,6 +556,10 @@ def setup_oscap():
     for daemon in ('foreman', 'httpd', 'foreman-proxy'):
         manage_daemon('restart', daemon)
 
+def oscap_content():
```
sthirugn: do we also need this function in fabfile.py if somebody wants to run it locally?

sthirugn:
ACK with non blocking comment

PR #331, opened 2016-05-05
Reorganized Upgrade on RHEVM

jyejare:
1. This PR reorganizes Upgrade Automation.
2. Same Description as PR#319 :
As planned earlier and due to non availability of provisioning on Openstack project we decided to move on rhevm. And so the reason we need to have functions for creating and deleting instances created from template on rhevm. This PR includes these two functions.

Note: Here in RHEVM template is same as openstack image.

While saving a template of any Sat server, one should save network, storage configuration and ssh-key with template otherwise this functions will fail to do an upgrade job for us.
I am still in research on how to add given shh key by automation.

So lets review the PR and Welcome RHEVM in Satellite Upgrade Automation.

Suggestions for improvements are always welcome.

PR #356, opened 2016-06-09
Removed old Upgrade code from automation_tools init file

jyejare:
This PR includes:
1. Remove old Upgrade Automation Tools code from automation_tools/__init__.py file as we have separate package/directory for python with its own modules.
2. The capsule upgrade command fix for capsule cert option which is now changed to --capsule-certs-tar for 6.2


elyezer:
Is fabfile needed to be updated?

ACK, pending question.

jyejare:
@elyezer OMG , Yes !

sthirugn:
ACK pending @elyezer's question

jyejare:
Nope, That doesnt requires any fabfile change. That change has been already done in PR #331.

PR #357, opened 2016-06-10
Set capsule Hostname env variable in both the cases

jyejare:
Set capsule Hostname fab env variable in both the cases:
1. When Given by Job User
2. Fetched from Jenkins Env Variables


sghai:
ACK.. merging it right away to test capsule upgrade..

PR #358, opened 2016-06-16
Capsule Upgrade Automation updates for 6.2 and Jenkins

jyejare:
This PR has updates for :
1. Some Capsule upgrade automation fixes and improvements for capsule RHEVM images
2. Updates for Capsule upgrade to 6.2
3. Jenkins related updates

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -133,6 +133,8 @@ def satellite6_capsule_upgrade(admin_password=None):
     TO_VERSION
         Capsule version to upgrade to and enable repos while upgrading.
         e.g '6.1','6.2'
+    CAPSULE_AK
+        AK name on Satellite using which capsule is registered
 
     """
     sat_host = env.get('satellite_host')
@@ -146,10 +148,27 @@ def satellite6_capsule_upgrade(admin_password=None):
         print('Wrong Capsule Version Provided to upgrade to. '
               'Provide one of 6.1, 6.2')
         sys.exit(1)
+    if not os.environ.get('CAPSULE_SUBSCRIPTION') and not os.environ.get(
+            'CAPSULE_AK'):
+        print('Neither CAPSULE_SUBSCRIPTION nor CAPSULE_AK environment '
+              'variables were defined. Make sure at least one of them is '
+              'defined.')
+        sys.exit(1)
     if admin_password is None:
         admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
     # Setting Capsule61 Repos
     major_ver = distro_info()[1]
+    # Re-register Capsule for 6.2
+    # AS per host unification feature: if there is a host registered where the
+    # Host and Content Host are in different organizations (e.g. host not in
+    # org, and content host in one), the content host will be unregistered as
+    # part of the upgrade process.
+    if to_version == '6.2':
+        ak_name = os.environ.get('CAPSULE_SUBSCRIPTION').split(',')[2].strip(
+            ) if os.environ.get('CAPSULE_SUBSCRIPTION') else os.environ.get(
+            'CAPSULE_AK')
+        run('subscription-manager register --org="Default_Organization" '
+            '--activationkey={0} --force'.format(ak_name))
     if os.environ.get('CAPSULE_URL') is None:
         enable_repos('rhel-{0}-server-satellite-capsule-{1}-rpms'.format(
             major_ver, to_version))
@@ -196,7 +215,7 @@ def satellite6_capsule_upgrade(admin_password=None):
             '/home/{0}-certs.tar'.format(cap_host))
     else:
         run('satellite-installer --scenario capsule --upgrade '
-            '--capsule-certs-tar /home/{0}-certs.tar'.format(cap_host))
+            '--certs-tar /home/{0}-certs.tar'.format(cap_host))
     print('CAPSULE UPGRADE finished at: {0}'.format(time.ctime()))
     # Test The status of all katello services
     run('katello-service status', warn_only=True)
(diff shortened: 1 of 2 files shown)


automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -146,10 +148,21 @@ def satellite6_capsule_upgrade(admin_password=None):
         print('Wrong Capsule Version Provided to upgrade to. '
               'Provide one of 6.1, 6.2')
         sys.exit(1)
+    if not os.environ.get('CAPSULE_SUBSCRIPTION') and not os.environ.get(
+            'CAPSULE_AK'):
+        print('One of Env. Var CAPSULE_SUBSCRIPTION or CAPSULE_AK should be'
```
elyezer: The if uses `and` operator and the message says `or`, what is the real intention here?

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -146,10 +148,21 @@ def satellite6_capsule_upgrade(admin_password=None):
...
+              'defined!')
+        sys.exit(1)
     if admin_password is None:
         admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
     # Setting Capsule61 Repos
     major_ver = distro_info()[1]
+    # Re-register Capsule for 6.2¬
```
elyezer: Is this a workaround? If yes, please incorporate the Bug ID to track where this comes from.

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -146,10 +148,21 @@ def satellite6_capsule_upgrade(admin_password=None):
...
+    if to_version == '6.2':
+        ak_name = os.environ.get('CAPSULE_SUBSCRIPTION')[2] if os.environ.get(
```
elyezer: Are you expecting `CAPSULE_SUBSCRIPTION` to be a list? The os.environ.get will return a string if it the env var is defined and getting the index 2 will actually returns its third character.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -402,10 +402,13 @@ def sync_capsule_tools_repos_to_upgrade(admin_password=None):
     run(initials + 'content-view publish --name {0} '
         '--organization {1}'.format(cv_name, org))
     # promote cv
-    lc_env_id = str(
+    lc_list = str(
         run(initials + 'lifecycle-environment list '
-            '--organization {0} | grep {1}'.format(org, env_name))).split(
-                '|')[0].strip()
+            '--organization {0} | grep {1}'.format(org, env_name))).split('\n')
+    for lc in lc_list:
```
elyezer: Why not grab the CSV output? You can use the csv module to parse the output and access the attributes by the column name.

elyezer:
ACK, pending comments.

automation_tools/satellite6/upgrade/__init__.py
sghai: AS per host unification feature: if there is a host registered where the Host and Content Host are in different organizations (e.g. host not in org, and content host in one), the content host will be unregistered as part of the upgrade process. 

And in 6.1, generally when we register capsule, org get associated with capsule content-host but not with host under "All Host". So this re-registration is required.
A thread is already started internally on list on same.

automation_tools/satellite6/upgrade/tasks.py
sghai: +1: parsing a CSV output is a cleaner way.

automation_tools/satellite6/upgrade/__init__.py
jyejare: The intention here is to provide either one env. var. If both are not given then the message will be printed to suggest either one env var to give and exits from the execution.

automation_tools/satellite6/upgrade/__init__.py
jyejare: Thanks @sghai. This is not a workaround but this is an impact of host unification feature.

automation_tools/satellite6/upgrade/__init__.py
jyejare: Yes. CAPSULE_SUBSCRIPTION is list of CV, ENV and AK details used for capsule subscription to be updated during upgrade. The details has to be provided in given sequence only.

elyezer: This needs to be updated, you need to process the string because all os.environ.get will return either a string or None if the env var was defined or if not respectively.

automation_tools/satellite6/upgrade/__init__.py
elyezer: What about making the message a little bit more clear? For example `Neither CAPSULE_SUBSCRIPTION nor CAPSULE_AK environment variables was defined. Define at least one of them in order to run this task.`

elyezer:
@sghai @jyejare thanks for the feedback, I left some reply comments where needed.

automation_tools/satellite6/upgrade/__init__.py
elyezer: The comment will be more informative if it states that the re-register is needed because the characteristic of how Satellite 6.1 register capsules.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -402,10 +402,13 @@ def sync_capsule_tools_repos_to_upgrade(admin_password=None):
     run(initials + 'content-view publish --name {0} '
         '--organization {1}'.format(cv_name, org))
     # promote cv
-    lc_env_id = str(
-        run(initials + 'lifecycle-environment list '
-            '--organization {0} | grep {1}'.format(org, env_name))).split(
-                '|')[0].strip()
+    lc_list = str(
+        run(initials + '--output csv lifecycle-environment list '
```
elyezer: You can also pass the name to the list command which will help filter even more the results:

```
[root@qe-sat6-rhel7 ~]# hammer lifecycle-environment list --organization-id 1 --name "Library"
[Foreman] Password for admin: 
---|---------|------
ID | NAME    | PRIOR
---|---------|------
1  | Library |      
---|---------|------
```

PR #360, opened 2016-06-21
 Update automation-tools to separate out target_image and hostname

kbidarkar:
a) This is required because if target_image and hostname remain same
then we would have the hostname for provisioning job as,
`qe-sat6-rhel7-base.satqe.lab.eng.rdu2.redhat.com`, which we do not
as it will cause DNS issues.
b) So separating target_image and hostname would enable us to have,
the hostname as `qe-sat6-rhel7.satqe.lab.eng.rdu2.redhat.com` and
the provisioning job could creae a virsh domain or image name with
`qe-sat6-rhel7-base`.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -645,6 +645,8 @@ def vm_create():
         base image name
     TARGET_IMAGE
         target image name
+    SERVER_HOSTNAME
+        server hostname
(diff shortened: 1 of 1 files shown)


rochacbruno:
ACK

sthirugn:
ACK

PR #362, opened 2016-06-29
Update the ping command for hosts which use ddns hash due to snapshot

kbidarkar:


lpramuk:
ACK. Looks good to me.

elyezer:
ACK

PR #363, opened 2016-06-29
Fix iso_download() as *SUM file format varies

### LINKED ISSUE #361: Fix iso_download() as *SUM file format varies
```<hash> *<iso_filename>```

vs.

```MD5 (<iso_filename>) = <hash>```

### PR DESCRIPTION
**Pull request by lpramuk**
Fix iso_download() as *SUM file format varies

- iso_download() fixed to match both SUM file patterns
- iso_install() fix if ISO already exists, and assumption that we run under root
- cdn_install() added 6.2 satellite rpm (usable for beta2)
- partition_disk() delete just '/home' (not 'home') from fstab
- product_install() we store SATELLITE_VERSION in local var but still it wasn't used
- updated docstrings (RHN_* env vars are not used by {iso,cdn}_install but by product_install)

(Fixes #361)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -273,8 +273,9 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
             print('Was not possible to fetch interface information')
             sys.exit(1)
 
-    proxy = 'capsule'
-    if os.environ.get('SATELLITE_VERSION') == '6.2':
+    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1'):
+        proxy = 'capsule'
+    else:
         proxy = 'foreman-proxy'
 
     installer_options = {
@@ -982,10 +983,10 @@ def downstream_install(admin_password=None, run_katello_installer=True):
     satellite_repo.close()
 
     # Install required packages for the installation
-    if os.environ.get('SATELLITE_VERSION') == '6.2':
-        run('yum install -y satellite')
-    else:
+    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1'):
         run('yum install -y katello')
+    else:
+        run('yum install -y satellite')
 
     installer_options = {
         'foreman-admin-password': admin_password,
@@ -1003,21 +1004,19 @@ def cdn_install(run_katello_installer=True):
 
     The following environment variables affect this command:
 
-    RHN_USERNAME
-        Red Hat Network username.
-    RHN_PASSWORD
-        Red Hat Network password.
-    RHN_POOLID
-        Optional. Red Hat Network pool ID. Determines what software will be
-        available from RHN.
     ADMIN_PASSWORD
         Optional, defaults to 'changeme'. Foreman admin password.
+    SATELLITE_VERSION
+        Satellite version.
 
     """
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
     # Install required packages for the installation
-    run('yum install -y katello')
+    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1'):
+        run('yum install -y katello')
+    else:
+        run('yum install -y satellite')
 
     installer_options = {
         'foreman-admin-password': admin_password,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by lpramuk** [comment]
I extended it a bit. 
Started by fixing iso install and fixed what came along my way.

File: automation_tools/__init__.py [hunk 1]
```diff
@@ -1003,21 +1003,19 @@ def cdn_install(run_katello_installer=True):
...
         Optional, defaults to 'changeme'. Foreman admin password.
+    SATELLITE_VERSION
+        Satellite version.
 
     """
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
     # Install required packages for the installation
-    run('yum install -y katello')
+    if os.environ.get('SATELLITE_VERSION') == '6.2':
```
elyezer: Since SATELLITE_VERSION is a required env var the task should fail it is not defined. Otherwise it will try to run katello whenever it is not defined and the user maybe be willing to install 6.2

automation_tools/__init__.py
```diff
@@ -1126,6 +1129,8 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     :param str certificate_url: where to fetch a fake certificate.
 
     """
+    # Fetch the Satellite Version information.
+    satellite_version = os.environ.get('SATELLITE_VERSION')
```
elyezer: Better to fail if the satellite_version is not defined and before doing any change on the server

automation_tools/__init__.py
```diff
@@ -1385,7 +1387,10 @@ def iso_download(iso_url=None):
                 quiet=True,
             )
             if result.succeeded:
-                iso_filename = search('\w+\s+\*?([^\s]+)', result).group(1)
+                # match either '<hash> *<iso_filename>'
+                # or '{MD5|SHA1|SHA256} (<iso_filename>) = <hash>'
+                iso_filename = search(
+                    '\w+\s+[\*\(]?([^\s\)]+)', result).group(1)
```
elyezer: Have validated this by running:

```
In [5]: s = ['hash *iso_filename.iso', 'MD5 (iso_filename.iso) = hash', 'SHA1 (iso_filename.iso) = hash', 'SHA256 (iso_filename.iso) = hash', ]

In [6]: for i in s:                                                                                                                            
    print(re.search('\w+\s+[\*\(]?([^\s\)]+)', i).group(1))
   ...:     
iso_filename.iso
iso_filename.iso
iso_filename.iso
iso_filename.iso
````

**Comment by elyezer** [comment]
ACK, pending failing earlier when a required env var is not defined.

File: automation_tools/__init__.py (hunk 3 above)
Comment (lpramuk) [inline automation_tools/__init__.py]: I also did in ipython. Thanks for this comment

File: automation_tools/__init__.py (hunk 1 above)
Comment (lpramuk) [inline automation_tools/__init__.py]: here it is similar with donwstream_install:
product_install -> downstream_install (requires SATELLITE_VERSION)
product_install -> cdn_install (requires SATELLITE_VERSION)

I better switch logic here if SATELLITE_VERSION not defined or 6.2 -> satellite
Now it is if not defined is the same as 6.1 -> katello

Comment (lpramuk) [inline automation_tools/__init__.py]: anyway it is ugly that product_install reads SATELLITE_VERSION and then runs {downstream|cdn}_install which both read SATELLITE_VERSION, param will be better

I'd like to address it in another PR.

File: automation_tools/__init__.py (hunk 2 above)
Comment (lpramuk) [inline automation_tools/__init__.py]: ```satellite_version``` assignment was already there I just shifted it to the beginning of product_install()
for downstream we are not failing anyway, better then fail is to assume 6.2 if not defined

**Comment by lpramuk** [comment]
I fixed it by reverse logic ifs:  if not defined then it is 6.2
```
    if os.environ.get('SATELLITE_VERSION') in ['6.0', '6.1']:
        run('yum install -y katello')
    else:
        run('yum install -y satellite')
```
We have to consider 6.2 as default and do special case for 6.1 ... not vice versa

lpramuk:
no need to fail when SATELLITE_VERSION is not defined
```os.environ.get('SATELLITE_VERSION')``` is then empty string and will be handled by else: clause

elyezer:
@lpramuk thanks for the clarifications.

ACK

rplevka:
ACK

PR #368, opened 2016-07-18
setup firewall rules before starting upgrade

sghai:
This is to ensure rules are properly set before running upgrade. We have already setup rules in rhev-m templates but for some reason they automatically resets. 

So these changes will ensure that rules are correctly placed.

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -11,7 +11,13 @@ from tasks import (
     sync_capsule_tools_repos_to_upgrade
 )
 from tools import copy_ssh_key, host_pings, reboot
-from automation_tools import foreman_debug, set_yum_debug_level, subscribe
+from automation_tools import (
+    foreman_debug,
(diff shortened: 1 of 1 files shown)


automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -184,6 +191,7 @@ def satellite6_capsule_upgrade(admin_password=None):
     print('Wait till packages update ... ')
     print('YUM UPDATE started at: {0}'.format(time.ctime()))
     update_packages(quiet=False)
+    setup_capsule_firewall()
```
jyejare: Better to put this above register command itself before starting any upgrade task on capsule.

Still, if you are ok here, put this after the below print line which says the yum update finished at <time>.

jyejare:
ACK pending comments.

automation_tools/satellite6/upgrade/__init__.py
sghai: +1, sure

PR #369, opened 2016-07-18
Disabling os update to investigate test slowness

sthirugn:


omaciel:
ACK

lpramuk:
I wonder what has been investigated?
Removing this means we are installing and testing Satellite on the same RHEL GA dot version on and on.
But customers have always latest (all errata applied) and documentation says so, so by this we diverged...

lpramuk:
Consider httpd or mongod erratum that could break things and now automation is not testing it and we will never catch it.

PR #372, opened 2016-07-20
refactored copy_ssh_key() to handle already existing priv rsa key

lpramuk:

### DIFF
--- a/automation_tools/satellite6/upgrade/tools.py
+++ b/automation_tools/satellite6/upgrade/tools.py
@@ -37,20 +37,25 @@ def copy_ssh_key(from_host, to_host):
     :param to_host: A string. Hostname on to which the ssh-key will be copied.
 
     """
+    execute(lambda: run('mkdir -p ~/.ssh'), host=from_host)
+    # do we have privkey? generate only pubkey
     execute(lambda: run(
-        '[ ! -f ~/.ssh/id_rsa.pub ] && '
-        'ssh-keygen -f ~/.ssh/id_rsa -t rsa -N \'\'', warn_only=True),
-        host=from_host)
-    if int(execute(lambda: run('[ -f ~/.ssh/id_rsa.pub ]; echo $?'),
-                   host=from_host)[from_host]) == 0:
-        pub_key = execute(lambda: run(
-            'cat ~/.ssh/id_rsa.pub'), host=from_host)[from_host]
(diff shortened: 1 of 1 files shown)


automation_tools/satellite6/upgrade/tools.py
```diff
@@ -37,20 +37,25 @@ def copy_ssh_key(from_host, to_host):
...
-        'ssh-keygen -f ~/.ssh/id_rsa -t rsa -N \'\'', warn_only=True),
-        host=from_host)
-    if int(execute(lambda: run('[ -f ~/.ssh/id_rsa.pub ]; echo $?'),
-                   host=from_host)[from_host]) == 0:
-        pub_key = execute(lambda: run(
-            'cat ~/.ssh/id_rsa.pub'), host=from_host)[from_host]
-        execute(lambda: run('[ ! -f ~/.ssh/authorized_keys ] && '
-                            'touch ~/.ssh/authorized_keys',
-                            warn_only=True), host=to_host)
+        '[ ! -f ~/.ssh/id_rsa ] || '
```
elyezer: ! -f means that the file does not exist, so accessing it will fail later

lpramuk: Short answer:
It shold read: It doesnt exist OR access it

Long answer:
! -f || cmd is equivalent to -f && cmd
The latter form return 1 as file doesnot exist and failing whole run() while the first form return 0 and run() is ok

elyezer: What I don't get is that why looking for `~/.ssh/id_rsa` instead of `~/.ssh/id_rsa.pub`

lpramuk: I'm looking for privkey to generate pubkey out of it 
if pubkey is present it is over written resulting in matching keypair (privkey could be from differrent keypait than pubkey)

lpramuk: Another approach would be to delete existing privkey and always generate keypair. 

But if I have image/instance with some important privkey I would delete it, so I chose approach to actually re-use existing privkey for our purpose...

elyezer: Fair enough

elyezer:
ACK, even though reading the `not or` logic is harder `and` the PR looks good.

jyejare:
ACK

PR #375, opened 2016-07-26
enable docker service in rhel7 after installing it

rplevka:
On RHEL7+ the `docker.service` is not being automatically enabled. Thus we need to enable it manually. Otherwise we'll end up with stopped `docker.service` everytime we spin up the system from a snapshot.


sghai:
ACK.

kbidarkar:
ACK

PR #385, opened 2016-08-12
If OS candidate repo was created then upgrade OS with it

lpramuk:
PR #369 disabled OS upgrade temporarily (permanently) and broke Vault testing run.
As test slowness investigation results are pending I realize we can still do OS update only if OS candidate repo is present.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1232,10 +1232,13 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         cdn_version=sat_cdn_version,
         host=host
     )
-    # Create custom OS beaker repo
+    # Create custom OS candidate repo and upgrade OS with it
     if os.environ.get('OS_UPGRADE_REPO'):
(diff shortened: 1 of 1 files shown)


sghai:
I think we should update `OS` from CDN only. Lets discuss with @sthirugn if we can enable it back instead of updating `OS` via `custom` repo.

lpramuk:
I used (and still would like to use) that custom repo for RHEL Vault testing. 
You can specify custom repo as beaker candidate RHEL-7.3-XXXXXX.X and voila you are testing Satellite on RHEL-7.3 OS.

So yes we can keep only one OS upgrade command but generally we have two things here:
- no custom repo specified -> plain OS update (errata applied out of CDN)
- custom candidate OS repo specified -> OS is being upgraded to candidate dot release.

lpramuk:
Lets revert PR #369 then, it wasn't a good idea...

sghai:
Ah.. I missed the case of handling Vault Testing..

lpramuk:
Reverting PR #387 is merged now, so no need for this PR

PR #387, opened 2016-08-16
Revert "Disabling os update to investigate test slowness"

lpramuk:
CDN is slow. Unless we have updated OS image on daily basis, we are required to run OS update always before installing Satellite.

We say: "Prior installing Satellite run yum update", so lets again do it in automation.


elyezer:
PR looks good, but since I am no aware of this should be merged or not I will leave for other folk to take a look.

ACK

omaciel:
ACK, I am not aware of any reasons why we need to disable this anymore.

PR #389, opened 2016-08-25
Setup firewalld only when installed

elyezer:
If firewalld is not installed fallback to iptables when setting up the
firewall

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -376,7 +376,8 @@ def setup_firewall(definitions=None, flush=True):
     if definitions is None:
         return
     os_version = distro_info()[1]
-    if os_version < 7:
+    has_firewalld = run('rpm -q firewalld', quiet=True).succeeded
+    if os_version < 7 or not has_firewalld:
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -376,7 +376,8 @@ def setup_firewall(definitions=None, flush=True):
     if definitions is None:
         return
     os_version = distro_info()[1]
-    if os_version < 7:
+    have_firewalld = run('rpm -q firewalld', quiet=True).succeeded
```
omaciel: s/have/has

omaciel:
ACK pending small nitpick

automation_tools/__init__.py
```diff
@@ -376,7 +376,8 @@ def setup_firewall(definitions=None, flush=True):
     if definitions is None:
         return
     os_version = distro_info()[1]
-    if os_version < 7:
+    has_firewalld = run('rpm -q firewalld', quiet=True).succeeded
+    if os_version < 7 or not has_firewalld:
```

PR #390, opened 2016-08-25
Install firewalld if it is not installed

elyezer:
Satellite 6 documentation encourages RHEL 7 user to use firewalld because that
install and use firewalld for RHEL 7 and above machines.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -376,8 +376,7 @@ def setup_firewall(definitions=None, flush=True):
     if definitions is None:
         return
     os_version = distro_info()[1]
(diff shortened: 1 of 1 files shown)


lpramuk:
ACK.

Thanks Elyezer for fixing my fix :-)

omaciel:
ACK

automation_tools/__init__.py
```diff
@@ -386,6 +385,8 @@ def setup_firewall(definitions=None, flush=True):
         if flush:
             run('iptables --flush')
     else:
+        if run('rpm -q firewalld', quiet=True).failed:
+            run('yum install -y firewalld')
```

PR #398, opened 2016-09-01
Enable plugins also in upstream

### LINKED ISSUE #391: Upstream installer does not install plugins which are included in Satellite
According to our upstream automation and `standalone installer` jobs, installed katello (satellite-upstream) doesn't have plugins installed (discovery, libvirt, etc.) which causes our tests to fail.

I believe we consider `upstream` Katello with all plugins which are bundled together and installed by default in Satellite 6 therefore I'm marking this issue as a bug

lpramuk:
Enable plugins also in upstream
- setup_default_capsule() is run also for upstream and zstream distribution
- product_install(): dont exlude upstream distro from postinstall plugins enablements
- upstream_install(): some pkgs needs to be installed explicitly in upstream

(Fixes #391)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -962,6 +962,18 @@ def upstream_install(
             sam='--sam' if sam else ''
         ))
 
+    # Install support for various compute resources in upstream
+    compute_resources = [
+        'ec2',
+        'gce',
+        'libvirt',
+        'openstack',
+        'ovirt',
+        'rackspace',
+        'vmware',
+    ]
+    run('yum install -y foreman-{{{0}}}'.format(','.join(compute_resources)))
+
     installer_options = {
         'foreman-admin-password': admin_password,
     }
@@ -1251,29 +1263,27 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         install_tasks[distribution], host=host, run_katello_installer=False
     )[host])
 
-    if distribution in (
-            'satellite6-cdn', 'satellite6-downstream', 'satellite6-iso'):
-        # When using VLAN Bridges os.environ.get('BRIDGE') is 'true' and
-        # executes with 'interface=eth0' for Satellite6-automation.
-        # When using Satellite6-installer one can specify custom interface.
-        if os.environ.get('BRIDGE') or os.environ.get('INTERFACE'):
-            # execute returns a dictionary mapping host strings to the given
-            # task's return value
-            installer_options.update(execute(
-                setup_default_capsule,
-                host=host,
-                # If an INTERFACE is specified it will be used otherwise would
-                # default to eth0 interface. Helpful for configuring DHCP and
-                # DNS capsule services.
-                interface=os.environ.get('INTERFACE') or 'eth0',
-                run_katello_installer=False
-            )[host])
-        else:
-            # execute returns a dictionary mapping host strings to the given
-            # task's return value.
-            installer_options.update(execute(
-                setup_default_capsule, host=host, run_katello_installer=False
-            )[host])
+    # When using VLAN Bridges os.environ.get('BRIDGE') is 'true' and
+    # executes with 'interface=eth0' for Satellite6-automation.
+    # When using Satellite6-installer one can specify custom interface.
+    if os.environ.get('BRIDGE') or os.environ.get('INTERFACE'):
+        # execute returns a dictionary mapping host strings to the given
+        # task's return value
+        installer_options.update(execute(
+            setup_default_capsule,
+            host=host,
+            # If an INTERFACE is specified it will be used otherwise would
+            # default to eth0 interface. Helpful for configuring DHCP and
+            # DNS capsule services.
+            interface=os.environ.get('INTERFACE') or 'eth0',
+            run_katello_installer=False
+        )[host])
+    else:
+        # execute returns a dictionary mapping host strings to the given
+        # task's return value.
+        installer_options.update(execute(
+            setup_default_capsule, host=host, run_katello_installer=False
+        )[host])
 
     # Firewall should be setup after setup_default_capsule clean the puppet
     # module it installs clean already created rules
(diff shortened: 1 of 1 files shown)


rplevka:
ACK

automation_tools/__init__.py
```diff
@@ -1322,18 +1322,17 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+            execute(setup_libvirt_key, host=host)
+        if satellite_version != '6.0':
+            execute(install_puppet_scap_client, host=host)
+        if satellite_version == '6.1':
+            execute(setup_oscap, host=host)
+        if satellite_version in ('6.2', ''):
+            execute(oscap_content, host=host)
+        if os.environ.get('PXE_DEFAULT_TEMPLATE_URL') is not None:
+            execute(setup_foreman_discovery, host=host)
+        execute(enable_ostree, host=host)
```
kbidarkar: Let's not change this unless we, have explicitly added support for  upstream.
There are features, which have it's own requirements, it would be better if the feature owner adds support for it in the upstream.

Example: oscap installs a package in downstream , but needs a puppet-module in upstream.

lpramuk: - satellite_version in ('6.2', '') - it doenst mean upstream it will be skipped anyway, I can add explicit condition to skip in upstream only this. I want the rest merged soon.


- default smart proxy isn't nice bonus?

lpramuk: Kedar, I will refactor it to meet your comments. thanks

lpramuk:
Final state of plugins/tasks for upstream: 
- setup_default_capsule enabled
- setup_libvirt_key enabled
- setup_foreman_discovery enabled
- ostree still disabled (will handle lpramuk in new PR)
- oscap still disabled (will handle kbidarka in new PR)

lpramuk:
cloned satellite6-installer pointed at my branch was successful installing upstream:
```
[hostname] Executing task 'setup_default_capsule'
...
[hostname] Executing task 'setup_libvirt_key'
...
[hostname] Executing task 'setup_foreman_discovery'
...
Done.
Disconnecting from root@hostname... done.
+ '[' false = true ']'
Deleting 6 temporary files
Started calculate disk usage of build
Finished Calculation of disk usage of build in 0 seconds
Started calculate disk usage of workspace
Finished Calculation of disk usage of workspace in 0 seconds
Finished: SUCCESS
```

rplevka:
ACK
looks good to me

automation_tools/__init__.py
```diff
@@ -1322,18 +1322,18 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     if distribution.startswith('satellite6'):
         execute(setup_default_docker, host=host)
         execute(katello_service, 'restart', host=host)
+        if os.environ.get('LIBVIRT_KEY_URL') is not None:
```
omaciel: Can we add a comment explaining the logic here? 3 months from now we want to remember why we have this condition :)

automation_tools/__init__.py
```diff
@@ -1322,18 +1322,18 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
             if satellite_version != '6.0':
                 execute(install_puppet_scap_client, host=host)
             if satellite_version == '6.1':
                 execute(setup_oscap, host=host)
             if satellite_version in ('6.2', ''):
                 execute(oscap_content, host=host)
-            if os.environ.get('PXE_DEFAULT_TEMPLATE_URL') is not None:
-                execute(setup_foreman_discovery, host=host)
             execute(enable_ostree, host=host)
+        if os.environ.get('PXE_DEFAULT_TEMPLATE_URL') is not None:
```
omaciel: Add a comment?

lpramuk: coz the task that is called uses that env var... ugly

we are going to refactor so that tasks will have params instead.

lpramuk: I'm going to add some explanation..er..comment

automation_tools/__init__.py
```diff
@@ -961,6 +961,8 @@ def upstream_install(
             os_version=os_version,
             sam='--sam' if sam else ''
         ))
+    run('yum install -y '
```
omaciel: This works but I was wondering if we should create a PLUGINS constant and list them there, in alphabetical order?

```python
PLUGINS = [
    'ec2',
    'gce', 
    'libvirt', 
    'openstack', 
    'ovirt', 
    'rackspace', 
    'vmware',
]

    run('yum install -y '
           'foreman-{{{0}}}'.format(','.join(PLUGINS))
```

omaciel:
@lpramuk none of my comments are blockers and can be either taken care of here or in a subsequent PR. ACK

automation_tools/__init__.py
lpramuk: I was thinking of it too as I saw it elsewhere in the code :)

omaciel:
Will come back later and will merge it as soon as @lpramuk adds one more change (he told me via irc that he would also handle my comment about plugins).

lpramuk:
Comments addressed.
- by adding comments
- by adding tabular list of compute resources to be installed for upstream

automation_tools/__init__.py
```diff
@@ -1251,29 +1263,27 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
-                interface=os.environ.get('INTERFACE') or 'eth0',
-                run_katello_installer=False
-            )[host])
-        else:
-            # execute returns a dictionary mapping host strings to the given
-            # task's return value.
-            installer_options.update(execute(
-                setup_default_capsule, host=host, run_katello_installer=False
-            )[host])
+    # When using VLAN Bridges os.environ.get('BRIDGE') is 'true' and
```
omaciel: Forgot to ask but why did you remove this conditional?

```diff
-    if distribution in (
 -            'satellite6-cdn', 'satellite6-downstream', 'satellite6-iso'):
```

kbidarkar: This is removed because we now want to configure, setup_default_capsule() for upstream, which was not being configured earlier.

kbidarkar:
ACK, all the suggestions and comments addressed.

automation_tools/__init__.py
```diff
@@ -962,6 +962,18 @@ def upstream_install(
...
+    compute_resources = [
+        'ec2',
+        'gce',
+        'libvirt',
+        'openstack',
+        'ovirt',
+        'rackspace',
+        'vmware',
+    ]
+    run('yum install -y foreman-{{{0}}}'.format(','.join(compute_resources)))
```

PR #399, opened 2016-09-01
BZ helper functions in provision-time

### LINKED ISSUE #394: Use robottelo's bz functions in provision-time
Sometimes we need to drive Satellite installation/postinstallation accordingly to bugzilla/redmine issue statuses.

For better picture I want to use such code in postinstall:
```
if os.environ.get('PXE_DEFAULT_TEMPLATE_URL') is not None:
    # for upstream skip until #16362 is fixed
    if not distribution.endswith('upstream') or not rm_bug_is_open(16362):
        execute(setup_foreman_discovery, host=host)
```

Lets discus how we can achieve it also in automation-tools. 
DRY vs. WET

lpramuk:
BZ helper functions
- added module bugzilla
- contains: bz_bug_is_open() and rm_bug_is_open()
- fixed logic of bz_bug_is_open()

(Fixes #394)

### DIFF
--- a/automation_tools/bugzilla.py
+++ b/automation_tools/bugzilla.py
@@ -1,0 +1,183 @@
+# -*- encoding: utf-8 -*-
+"""
+Collection of functions to work with Bugzilla and Redmine.
+copied from robottelo's robotello/decorators/__init__.py
+"""
+import bugzilla
+import logging
+import requests
+
+from six.moves.xmlrpc_client import Fault
+from xml.parsers.expat import ExpatError, ErrorString
+
+BZ_OPEN_STATUSES = [
+    'NEW',
+    'ASSIGNED',
+    'POST',
+    'MODIFIED'
+]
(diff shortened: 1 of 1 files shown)


elyezer:
I am not sure about duplicating code for the same task here and on robottelo... maybe a separated project/repo would be better...

rochacbruno:
@elyezer there was some discussion in https://github.com/SatelliteQE/automation-tools/issues/394#issuecomment-243431755 robottelo uses some settings from properties file, and looks like it is already relying on external lib.

If we going to create a new project/lib to hold that we will need to make it flexible to handle settings from both scenarios.

elyezer:
If here is a simpler version then more clean up should be done, for example we have comments mentioning `skip_if_rm_bug_open` which is not present here for example.

lpramuk:
Resolved comments. Most probably :-)

elyezer:
ACK

rochacbruno:
ACK

We decided to merge it, and in future, if a lot of changed needed here we are going to make it an external library for use in both projects.

PR #402, opened 2016-09-06
Enable ostree plugin also for upstream

### LINKED ISSUE #391: Upstream installer does not install plugins which are included in Satellite

lpramuk:
Enable ostree plugin also for upstream
- enable_ostree() refactored to handle plugin enablement also in upstream

(Fixes #391)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -604,17 +604,23 @@ def setup_foreman_discovery():
     run('rm -rf {0}'.format(template_file))
 
 
-def enable_ostree():
-    """Task to enable ostree plugin for Satellite on rhel7. This enables ostree
-    type repository.
+def enable_ostree(upstream=False):
+    """Task to enable ostree plugin for Satellite on rhel7.
+    This enables ostree type repository.
 
+    :param bool upstream: Flag whether we run on upstream. Default: ``False``.
     """
     os_version = distro_info()[1]
-    sat_version = os.environ.get('SATELLITE_VERSION')
-    # Enable ostree plugin only for Satellite6.2 installed on rhel7
-    if sat_version == '6.2' and os_version >= 7:
-        run('satellite-installer --scenario satellite '
-            '--katello-enable-ostree=true')
+    if os_version >= 7:
+        if upstream:
+            create_custom_repos(centos_atomic='http://buildlogs.centos.org/'
+                                'centos/7/atomic/x86_64/Packages/')
+        run('{}-installer --scenario {} --katello-enable-ostree=true'
+            .format(*['foreman', 'katello'] if upstream else ['satellite']*2))
+        if upstream:
+            delete_custom_repos('centos_atomic')
+    else:
+        print('ostree plugin is supported only on rhel7+')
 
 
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -1342,7 +1348,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
                 execute(setup_oscap, host=host)
             if satellite_version in ('6.2', ''):
                 execute(oscap_content, host=host)
-            execute(enable_ostree, host=host)
+        # ostree plugin is for Sat6.2+ rhel7 only
+        if satellite_version not in ('6.0', '6.1'):
```
omaciel: @lpramuk you'll have to rebase this PR, but as you do that, consider whether the `enable_ostree` task should be executed on `downstream`, `upstream` or both?

sghai: Should run on both but only on rhel7

automation_tools/__init__.py
```diff
@@ -604,17 +604,23 @@ def setup_foreman_discovery():
...
     """
     os_version = distro_info()[1]
-    sat_version = os.environ.get('SATELLITE_VERSION')
-    # Enable ostree plugin only for Satellite6.2 installed on rhel7
-    if sat_version == '6.2' and os_version >= 7:
-        run('satellite-installer --scenario satellite '
-            '--katello-enable-ostree=true')
+    if os_version >= 7:
+        if upstream:
+            create_custom_repos(Centos_Atomic='http://buildlogs.centos.org/'
```
sghai: `s/Centos_Atomic/centos_atomic` ?

automation_tools/__init__.py
```diff
@@ -604,17 +604,23 @@ def setup_foreman_discovery():
...
+                                'centos/7/atomic/x86_64/Packages/')
+        run('{}-installer --scenario {} --katello-enable-ostree=true'
+            .format(*['foreman', 'katello'] if upstream else ['satellite']*2))
+        if upstream:
+            delete_custom_repos('Centos_Atomic')
```
sghai: `s/Centos_Atomic/centos_atomic` ?

automation_tools/__init__.py
lpramuk: not in ('6.0', '6.1') => 6.2 and nightly

automation_tools/__init__.py
lpramuk: param name is translated to repo name.

lpramuk:
Comments resolved.
Rebased.

Tempted to reverse logic for oscap_content task in favor of smarter 'not in old versions' to avoid adding new versions with every new release (6.3, 6.4...)

sghai:
ACK

rplevka:
ACK

PR #404, opened 2016-09-08
Changes to setup discovery7.0 in upstream

sghai:
- Added a new task to setup discovery for upstream `setup_foreman_discovery_70`. 
- Please note that steps to setup discovery in upstream is different than 6.1, 6.2 releases. So I proffered to define separate task and same would be applicable for upcoming 6.3 release.

Please review it and suggest if any changes are required.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -573,31 +573,75 @@ def install_puppet_scap_client():
     run('yum -y install puppet-foreman_scap_client', warn_only=True)
 
 
-def setup_foreman_discovery():
-    """Task to setup foreman discovery."""
-    # Install required packages to enable discovery plugin
+def setup_foreman_discovery(sat_version, upstream=False):
+    """Task to setup foreman discovery.
+
+    The following environment variables affect this task:
+
+    * `PXE_DEFAULT_TEMPLATE_URL`
+    * `PXE_DEFAULT_TEMPLATE_URL_FOR_70`
+    * `PXELINUX_DISCOVERY_SNIPPET_URL`
+
+    :param str sat_version: should contain satellite version e.g. 6.1, 6.2
+    :param bool upstream: Flag whether we run on upstream. Default: ``False``.
+    """
     packages = (
         'tfm-rubygem-foreman_discovery',
         'rubygem-smart_proxy_discovery',
-        'foreman-discovery-image',
         'tfm-rubygem-hammer_cli_foreman_discovery',
     )
-    run('yum install -y {0}'.format(' '.join(packages)))
+    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
+
+    if not upstream and sat_version in ('6.1', '6.2'):
+        run('yum install -y {0}'.format(' '.join(packages)), warn_only=True)
+        run('yum install -y foreman-discovery-image')
+        for daemon in ('foreman', 'httpd', 'foreman-proxy'):
+            manage_daemon('restart', daemon)
+        template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
+        template_file = run('mktemp')
+        hostname = run('hostname -f', quiet=True).strip()
+        run('wget -O {0} {1}'.format(template_file, template_url))
+        run('sed -i -e "s/SatelliteCapsule_HOST/{0}/" {1}'
+            .format(hostname, template_file))
+        # Update the template
+        run('hammer -u admin -p {0} template update --name '
+            '"PXELinux global default" --file {1}'
+            .format(admin_password, template_file))
+        run('rm -rf {0}'.format(template_file))
+        return
+
+    if upstream:
+        run('yum install -y {0}'.format(' '.join(packages)), warn_only=True)
+        # Fetch the nightly FDI from upstream
+        image_url = 'http://downloads.theforeman.org/discovery/nightly/fdi-image-latest.tar'  # noqa
+        run('wget {0} + -O - | tar x --overwrite -C /var/lib/tftpboot/boot'
+            .format(image_url))
+    if not upstream and sat_version == '6.3':
+        # In 6.3, installer should install all required packages except FDI
+        run('rpm -q {0}'.format(' '.join(packages)))
+        run('yum install -y foreman-discovery-image')
 
     for daemon in ('foreman', 'httpd', 'foreman-proxy'):
-        manage_daemon('restart', daemon)
-
-    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
-    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
-    if template_url is None:
-        print('You must specify the PXE default template URL')
-        sys.exit(1)
-
+            manage_daemon('restart', daemon)
+    # Unlock the default Locked template for discovery
+    run('hammer -u admin -p {0} template update --name '
+        '"PXELinux global default" --lock "false"'
+        .format(admin_password))
+    # Fetch the updated template where ONTIMEOUT set to 'discovery'
+    # Note that this template is for discovery7.0
+    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL_FOR_70')
+    # Fetch the updated discovery snippet where URL includes proxy port
+    snippet_url = os.environ.get('PXELINUX_DISCOVERY_SNIPPET_URL')
+    snippet_file = run('mktemp')
+    run('wget -O {0} {1}'.format(snippet_file, snippet_url))
+    # Update the template
+    run('hammer -u admin -p {0} template update --name '
+        '"pxelinux_discovery" --file {1}'
+        .format(admin_password, snippet_file))
+    run('rm -rf {0}'.format(snippet_file))
     template_file = run('mktemp')
-    hostname = run('hostname -f', quiet=True).strip()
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1348,13 +1405,23 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+                # if we have PXE default template we can setup discovery
+                if os.environ.get('PXE_DEFAULT_TEMPLATE_URL') is not None:
+                    # setup_foreman_discovery is applicable for 6.1 and 6.2
+                    execute(setup_foreman_discovery, host=host)
         # ostree plugin is for Sat6.2+ and nightly (rhel7 only)
         if satellite_version not in ('6.0', '6.1'):
             execute(
                 enable_ostree, distribution.endswith('upstream'), host=host)
+        # setup_foreman_discovery_70 is for upstream and upcoming 6.3 release
+        if satellite_version not in ('6.0', '6.1', '6.2'):
```
kbidarkar: I think if we need this on priority, for now the check needs to be `distribution.endswith('upstream') or satellite_version == '6.3'`

As currently satellite_version even for upstream is '6.2'

automation_tools/__init__.py
```diff
@@ -604,6 +604,63 @@ def setup_foreman_discovery():
     run('rm -rf {0}'.format(template_file))
 
 
+def setup_foreman_discovery_70(upstream=False):
```
kbidarkar: I think we should merge `setup_foreman_discovery` and `setup_foreman_discovery_70` into one method.
We can pass in 2 args , sat_version and upstream to control the flow, in that single method, this way may be can reduce the unwanted code.

automation_tools/__init__.py
sghai: Yeah.. completely forgot about `6.3`. Since we need this on priority so will make changes as per your suggestion.

automation_tools/__init__.py
sghai: I think only package installer part is being duplicated. But it would be easier to manage in separate setups. I'll still give this a try and combine them.

omaciel: Agreed with @kbidarkar :)

automation_tools/__init__.py
```diff
@@ -1348,13 +1405,23 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+                if os.environ.get('PXE_DEFAULT_TEMPLATE_URL') is not None:
+                    # setup_foreman_discovery is applicable for 6.1 and 6.2
+                    execute(setup_foreman_discovery, host=host)
         # ostree plugin is for Sat6.2+ and nightly (rhel7 only)
         if satellite_version not in ('6.0', '6.1'):
             execute(
                 enable_ostree, distribution.endswith('upstream'), host=host)
+        # setup_foreman_discovery_70 is for upstream and upcoming 6.3 release
+        if distribution.endswith('upstream') or satellite_version == '6.3':
+            if os.environ.get('PXE_DEFAULT_TEMPLATE_URL_FOR_70') is not None:
```
kbidarkar: Also, I think we can get rid of these ENV_VAR checks, and specify the needed ENV_VAR explicitly for that method inside the method itself. I feel as we ourselves handle these ENV_VAR in jenkins config_files there is no need to check these.
If anyone using the method will know , which ENV_VAR to specify from the method, like we have for other methods currently.

Just a thought. :)

sghai:
@kbidarkar : Ready for another round. Please take a look.

automation_tools/__init__.py
```diff
@@ -573,31 +573,68 @@ def install_puppet_scap_client():
...
+        run('rm -rf {0}'.format(template_file))
+        return
 
-    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
-    template_url = os.environ.get('PXE_DEFAULT_TEMPLATE_URL')
-    if template_url is None:
-        print('You must specify the PXE default template URL')
-        sys.exit(1)
+    if upstream and sat_version == '6.2':
+        run('yum install -y {0}'.format(' '.join(packages)), warn_only=True)
```
lpramuk: I would omit sat_version= 6.2 here. (it's nonsense exclusive options with and)

automation_tools/__init__.py
```diff
@@ -573,31 +573,68 @@ def install_puppet_scap_client():
...
         'rubygem-smart_proxy_discovery',
-        'foreman-discovery-image',
         'tfm-rubygem-hammer_cli_foreman_discovery',
     )
-    run('yum install -y {0}'.format(' '.join(packages)))
+    admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
-    for daemon in ('foreman', 'httpd', 'foreman-proxy'):
-        manage_daemon('restart', daemon)
+    if sat_version in ('6.1', '6.2'):
```
lpramuk: You wanted to exclude upstream here instead of including upstream in next if (see next comment)
```if not upstream and sat_version in ('6.1', '6.2'):```

Comment (lpramuk) [inline automation_tools/__init__.py]: We have to fix our framework so that it cannot happen that ```upstream=True and sat_version in ('6.2', '6.1',...)```

File: automation_tools/__init__.py [hunk 6]
```diff
@@ -573,31 +573,68 @@ def install_puppet_scap_client():
...
-    if template_url is None:
-        print('You must specify the PXE default template URL')
-        sys.exit(1)
+    if upstream and sat_version == '6.2':
+        run('yum install -y {0}'.format(' '.join(packages)), warn_only=True)
+        # Fetch the nightly FDI from upstream
+        image_url = 'http://downloads.theforeman.org/discovery/nightly/fdi-image-latest.tar'  # noqa
+        run('wget {0} + -O - | tar x --overwrite -C /var/lib/tftpboot/boot'
+            .format(image_url))
+    if sat_version == '6.3':
```
lpramuk: Same here ```if not upstream and sat_version == '6.3':```

sghai:
@lpramuk : thank you for comments. PR is updated.

sghai:
Need one more reviewer ? @kbidarkar ? Please take another look.

automation_tools/__init__.py
```diff
@@ -573,31 +573,69 @@ def install_puppet_scap_client():
...
+        # Fetch the nightly FDI from upstream
+        image_url = 'http://downloads.theforeman.org/discovery/nightly/fdi-image-latest.tar'  # noqa
+        run('wget {0} + -O - | tar x --overwrite -C /var/lib/tftpboot/boot'
+            .format(image_url))
+    if not upstream and sat_version == '6.3':
+        # In 6.3, installer should install all required packages except FDI
+        run('rpm -q {0}'.format(' '.join(packages)))
+        run('yum install -y foreman-discovery-image')
 
+    for daemon in ('foreman', 'httpd', 'foreman-proxy'):
```
lpramuk: lines 618 - 640 seems common for 6.3 and for upstream but not for 6.1 and 6.2 
so you wanted to indent these line not be executed twice for 6.1 and 6.2

so you have to join two main ifs into one:
```
if upstream or sat_version == '6.3':
```
and add smaller nested ifs to exlude only for upstream/6.3

automation_tools/__init__.py
```diff
@@ -573,31 +573,69 @@ def install_puppet_scap_client():
...
+    run('rm -rf {0}'.format(snippet_file))
     template_file = run('mktemp')
-    hostname = run('hostname -f', quiet=True).strip()
     run('wget -O {0} {1}'.format(template_file, template_url))
-    run('sed -i -e "s/SatelliteCapsule_HOST/{0}/" {1}'
-        .format(hostname, template_file))
+    # Update the template
     run('hammer -u admin -p {0} template update --name '
         '"PXELinux global default" --file {1}'
         .format(admin_password, template_file))
```
lpramuk: ....up to this line (indented lines)

automation_tools/__init__.py
sghai: for 6.1 and 6.2, I added a `return` statement. As if that condition sets then after performing all steps, control exits.

lpramuk: I overlooked it. then ok

lpramuk:
ACK.

automation_tools/__init__.py
```diff
@@ -573,31 +573,69 @@ def install_puppet_scap_client():
     run('yum -y install puppet-foreman_scap_client', warn_only=True)
 
 
-def setup_foreman_discovery():
-    """Task to setup foreman discovery."""
-    # Install required packages to enable discovery plugin
+def setup_foreman_discovery(sat_version, upstream=False):
+    """Task to setup foreman discovery.
+
```
kbidarkar: Let's add the ENV_VAR required for `setup_foreman_discovery`, similar to `setup_ddns` method, it would help us quickly verify the required ones.
`The following environment variables affect this command:
  * `PXE_DEFAULT_TEMPLATE_URL_FOR_70`
  * `PXELINUX_DISCOVERY_SNIPPET_URL`

I know it's not available for all the methods, but I feel it is better to have this.

sghai:
thanks @kbidarkar. Updated ENV_VAR. And Jenkins is all set with these variables too.

sghai:
@kbidarkar could you please take a look if all your comments are addressed ?

kbidarkar:
@sghai , sorry missed to review.

kbidarkar:
ACK, LGTM.

PR #407, opened 2016-09-13
Fix: use sat6qe/katellovirt instead

lpramuk:
Fix: use sat6qe/katellovirt instead
- domcleal/katellovirt (unmaintained) was removed from puppetforge


kbidarkar:
ACK

PR #411, opened 2016-09-16
Fixes issues with flake8 and install satellite via repo_url and ak.

kbidarkar:
1) Fixes flake8 issues of PR 408

2) Updated the logic of PR 408, to subscribe for all distribution
methods and then unsubscribe for just 'satellite6-activationkey'.

3) Also now updating `distribution` instead of
`install_tasks[distribution]`

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1088,7 +1088,9 @@ def downstream_install(admin_password=None, run_katello_installer=True):
     else:
         return installer_options
 
-def repofile_install(admin_password=None, run_katello_installer=True, repo_url=None):
+
+def repofile_install(admin_password=None, run_katello_installer=True,
+                     repo_url=None):
     """Task to install Satellite 6.3 via repo files
 
     The following environment variables affect this command:
@@ -1103,7 +1105,7 @@ def repofile_install(admin_password=None, run_katello_installer=True, repo_url=N
         admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
     if repo_url is None:
-       repo_url = os.environ.get('REPO_FILE_URL')
+        repo_url = os.environ.get('REPO_FILE_URL')
 
     run('yum install -y wget')
     run('wget -O /etc/yum.repos.d/satellite63.repo {0}'.format(repo_url))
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -1341,12 +1347,19 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
+        'satellite6-activationkey': ak_install,
         'satellite6-iso': iso_install,
         'satellite6-upstream': upstream_install,
     }
     distribution = distribution.lower()
+
+    # Make sure downstream_install only is called for sat6.1 and sat6.2, as
+    # repofile_install and ak_install can be passed only for Sat6.3.
+    if satellite_version in ('6.1', '6.2'):
+        if distribution in ('satellite6-repofile', 'satellite6-activationkey'):
```
omaciel: Should this be `not in`?

automation_tools/__init__.py
```diff
@@ -1421,11 +1434,11 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     if distribution in ('satellite6-downstream', 'satellite6-iso'):
         execute(java_workaround, host=host)
 
-    # Make sure downstream_install only is called for sat6.1 and sat6.2, as
-    # repofile_install and ak_install can be passed only for Sat6.3.
-    if satellite_version in ('6.1','6.2'):
-        if install_tasks[distribution] in ('repofile_install', 'ak_install'):
-            install_tasks[distribution] = 'downstream_install'
+    # Do not subscribe to cdn for Satellite6.3 if using Activation Keys of
```
omaciel: You comment is a bit confusing to me. 'Do not subscribe' means that the system is already subscribed, but you execute an `unsubscribe` task. I think what you want to say is:

```python
# If using the internal dog food server, unsubscribe from `CDN` first.
```

kbidarkar: Ok, agreed, it could be confusing, will update as suggested.

automation_tools/__init__.py
kbidarkar: Instead of calling repofile_install and activationkey_install for satellite_version 6.1 and 6.2, I want it to default to downstream_install. Hence I think `distribution in` is fine.

omaciel:
ACK

PR #415, opened 2016-09-19
Added server-optional-rpms repo for installing grub2-efi-modules.

kbidarkar:
a) Will be adding this only for repofile_install

b) Installation with ak should not require enabling this and should
   be set to default via the dogfoods ak.


kbidarkar:
Damn, I had read @sghai mail but missed these things.

sthirugn:
ACK

sthirugn:
Merging with one ACK since @kbidarkar is testing 6.3 trigger now.

PR #418, opened 2016-09-20
fixed the template type in discovery setup

sghai:
--type option was missing in template cmd.


kbidarkar:
Will merge this for now.

Issue #416, opened 2016-09-19
add optional-rpms to repofile_install for rhel6


kbidarkar:
currently the repo appears to exists for RHEL7 only.

Track/Investigate and add support for RHEL6 too.

lpramuk:
# subscription-manager repos --enable rhel-*-server-optional-rpms
Repository 'rhel-6-server-optional-rpms' is enabled for this system.

kbidarkar:
Looks like the actual stuff that we need here is installation of `grub2-efi-modules` package for optional-rpms from rhel6. Last time I checked, I couldn't find the repo, had to investigate this further.

kbidarkar:
Discussions and explanation for this exists in PR #425

PR #425, opened 2016-09-27
[Do Not Merge] Fix repo_install as RHEL6 has also optional channel available

lpramuk:
Fix repo_install as RHEL6 has also optional channel available
(...if optional is not available then entitlements are about to fix)

(Issue #416)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1110,9 +1110,8 @@ def repofile_install(admin_password=None, run_katello_installer=True,
     run('wget -O /etc/yum.repos.d/satellite63.repo {0}'.format(repo_url))
 
     # Enable required repository
-    if os_version >= 7:
-        run('subscription-manager repos --enable "rhel-{0}-server-optional-rpms"'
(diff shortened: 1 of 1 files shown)


kbidarkar:
Are you able to install grub2-efi package from rhel6 optional-rpms repo ?

If not, do you know from which repo we can find it out ?

kbidarkar:
The exact package name is `grub2-efi-modules`.

lpramuk:
rhel-6-server-optional-rpms doesnt have it while rhel-7-server-optional-rpms has:
```grub2-efi-modules.x86_64                         1:2.02-0.34.el7_2                          @rhel-7-server-optional-rpms```

lpramuk:
so it has to be contained by Satellite compose then (on rhel6)

lpramuk:
but I dont see currently foreman, foreman-plugins for rhel6
Thus we dont have to solve it (at least for now)

kbidarkar:
@lpramuk , thanks for investigating into this task.

PR #427, opened 2016-09-27
enable_ostree now uses sat_version to control the flow

kbidarkar:
a) depending upon the satellite version automatically handle the
   disable-system-checks.
b) Now the disable-system-checks installer option is only provided
   for sat6.3
c) This also will as per the plan to keep only sat_version to
   control the flow instead of having both distribution and
   sat_version, after the jenkins reorg.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -654,7 +654,7 @@ def setup_foreman_discovery(sat_version, upstream=False):
     run('rm -rf {0}'.format(template_file))
 
 
-def enable_ostree(upstream=False):
+def enable_ostree(sat_version='6.3'):
     """Task to enable ostree plugin for Satellite on rhel7.
     This enables ostree type repository.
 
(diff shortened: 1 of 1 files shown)


JacobCallahan:
ACK

automation_tools/__init__.py
```diff
@@ -654,21 +654,24 @@ def setup_foreman_discovery(sat_version, upstream=False):
...
                                 'centos/7/atomic/x86_64/Packages/')
-        run('{}-installer --scenario {} --disable-system-checks '
-            '--katello-enable-ostree=true'
-            .format(*['foreman', 'katello'] if upstream else ['satellite']*2))
-        if upstream:
+        run('{0}-installer --scenario {1} {2} --katello-enable-ostree=true'
+            .format(
+            'foreman' if sat_version=='nightly' else 'satellite',
+            'katello' if sat_version=='nightly' else 'satellite',
+            '--disable-system-checks' if sat_version=='6.3' else '')
```
omaciel: This could potentially cause issues when we hit `6.4`...

kbidarkar: I think I shall handle it then for that too. :)

PR #428, opened 2016-09-27
Fix pep8 and added common Make targets (travis enabled)

### LINKED ISSUE #426: Enable flake8 run via travis for automation-tools
I want to focus on code rather than remembering to run this each time I change something, if travis build fails, would like to fix it ASAP.

rochacbruno:
Closes #426 

- @omaciel enabled travis
- Added Makefile common targets
- Added commit hook to check flake8 before commiting `make install-commit-hook` will do that

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,40 @@
+# Commands --------------------------------------------------------------------
+
+help:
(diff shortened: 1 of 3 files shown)


kbidarkar:
ACK, really nice to have this, helps a lot.

Issue #422, opened 2016-09-27
Fix automation-tools to use satellite_version==nightly , rather than distribution==upstream


kbidarkar:
1) Fix all the instances of distribution==upstream with satellite_version==nightly 
2) This will help with consistency

lpramuk:
I'm also in favor of ```satellite_version == 'nightly'``` instead of ```distribution == 'upstream'

by ```satellite_version``` we say WHAT to install (6.2 6.3 nightly)
by ```distribution``` we say the WAY to install (internal, cdn, repofile...)

kbidarkar:
Fixed in , https://github.com/SatelliteQE/automation-tools/pull/431

PR #431, opened 2016-10-06
Handle logic of methods using sat_version instead of two variables.

kbidarkar:
a) Earlier we had both sat_versions and upstream variables to
   handle the flow of logic.
b) Now all the logic flow will be depending upon the sat_version
   variable.
c) This will reduce complexity and simplify things.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -579,7 +579,7 @@ def install_puppet_scap_client():
     run('yum -y install puppet-foreman_scap_client', warn_only=True)
 
 
-def setup_foreman_discovery(sat_version, upstream=False):
+def setup_foreman_discovery(sat_version):
     """Task to setup foreman discovery.
 
     The following environment variables affect this task:
@@ -589,7 +589,6 @@ def setup_foreman_discovery(sat_version, upstream=False):
     * `PXELINUX_DISCOVERY_SNIPPET_URL`
 
     :param str sat_version: should contain satellite version e.g. 6.1, 6.2
-    :param bool upstream: Flag whether we run on upstream. Default: ``False``.
     """
     packages = (
         'tfm-rubygem-foreman_discovery',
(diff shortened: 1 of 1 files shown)


kbidarkar:
I could only find those instances , which required update. If there are any other instances for which we need to remove the upstream check, will do so. :)

automation_tools/__init__.py
```diff
@@ -1545,24 +1545,21 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
-                execute(setup_oscap, host=host)
-            if satellite_version not in ('6.0', '6.1'):
-                execute(oscap_content, host=host)
+        if satellite_version != '6.0':
+            execute(install_puppet_scap_client, host=host)
+        if satellite_version == '6.1':
+            execute(setup_oscap, host=host)
+        if satellite_version not in ('6.0', '6.1'):
+            execute(oscap_content, host=host)
         # ostree plugin is for Sat6.2+ and nightly (rhel7 only)
```
lpramuk: lines 1554 to 1559 either needs to be indented into one common ```if``` or yet better now all 3 ifs can merged with logic sat version != nightly
1. ```if satellite_version not in ('6.0', 'nightly'):``` (added or not nightly)
2. ```if satellite_version == '6.1':``` (stays same)
3. ```if satellite_version not in ('6.0', '6.1', 'nightly'):``` (added or not nightly)

lpramuk:
Implements issue #422

automation_tools/__init__.py
kbidarkar: I wanted to avoid an indent , so removed it but missed to add back nightly checks in the if.

kbidarkar:
@lpramuk , if you were referring to ostree plugin it does work for nightly.

sghai:
ACK.. Looks good.

sghai:
@lpramuk : Please feel free to merge if you'r fine with changes..

PR #432, opened 2016-10-13
updated order of execution of setup_discovery_task

sghai:
Y'day while working with @oshtaier I found that 'PXELinux global default' template is not being updated via automation. I was discussing with @kbidarkar and we realized this could be due to re-running the installer after the setup discovery task. Thanks Kedar for tip.
 
We are running `enable_ostree` task immediately after `setup_discovery_task`.  And enable_ostree task re-run the installer and that leads to remove any custom changes to template. 

So to fix this I just moved the `setup_discovery_task` after the `enable_ostree` task and hopefully this will fix the issue with all discovered host failures.


oshtaier:
ACK

kbidarkar:
ACK, glad it works now.

PR #435, opened 2016-10-28
Fix for making firewall persistent across reboot for RHEL6

kbidarkar:
a) We never had firewall persistent across reboot for RHEL6
b) We were just saving the firewall rules.
c) iptables service was never set to start at boot time.
d) With this fix, we can now save the iptables using he service
   command and also enable it too.


lpramuk:
ACK

PR #436, opened 2016-11-03
Do not subscribe to CDN when using an ActivationKey

omaciel:
When you install a Satellite system using an ActivationKey from a host Satellite 6, do not subscribe to CDN. All content and prerequisite packages should be installed from the host Satellite 6 instead.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1173,7 +1173,6 @@ def ak_install(admin_password=None, clean_beaker=True,
 
     # Clean up and install with basic packages.
     clean_rhsm()
-    update_basic_packages()
 
     # Install the cert file
     run('yum -y localinstall {0}/pub/katello-ca-consumer-latest.noarch.rpm'
@@ -1186,10 +1185,21 @@ def ak_install(admin_password=None, clean_beaker=True,
         '--activationkey="{1}"'.format(org, act_key)
     )
     run(cmd)
+
+    # Refresh subscriptions and clean up YUM
     print('Refreshing Subscription-manager.')
     run('subscription-manager refresh')
     print('Performing yum clean up.')
     run('yum clean all', quiet=True)
+
+    # Set yum stdout log level to be less verbose
+    set_yum_debug_level()
+
+    # Install some basic packages
+    update_basic_packages()
+    install_prerequisites()
+    # Update the machine
+    update_packages()
 
     # Install required packages for the installation
     run('yum install -y satellite')
@@ -1434,14 +1444,20 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     if distribution == 'satellite6-cdn' and test_in_stage:
         execute(update_rhsm_stage, host=host)
 
-    # Subscribe for all distributions to install prerequisities and updates.
-    # Then let's unsubscribe just before calling ak_install.
-    execute(subscribe, host=host)
+    # Create custom OS beaker repo
+    if os.environ.get('OS_UPGRADE_REPO'):
+        os_upgrade_repo = os.environ.get('OS_UPGRADE_REPO')
+        execute(create_custom_repos, rhel_candidate=os_upgrade_repo, host=host)
 
-    # Setting yum stdout log level to be less verbose
-    execute(set_yum_debug_level, host=host)
-
-    execute(install_prerequisites, host=host)
+    # If NOT using an activationkey, subscribe to CDN and install some
+    # basic required packages.
+    if distribution != 'satellite6-activationkey':
+        execute(subscribe, host=host)
+        execute(install_prerequisites, host=host)
+        # Setting yum stdout log level to be less verbose
+        execute(set_yum_debug_level, host=host)
+        # Update the machine
+        execute(update_packages, host=host, warn_only=True)
 
     execute(setenforce, selinux_mode, host=host)
     execute(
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1173,7 +1173,6 @@ def ak_install(admin_password=None, clean_beaker=True,
 
     # Clean up and install with basic packages.
     clean_rhsm()
-    update_basic_packages()
```
omaciel: Since the system is not yet registered, we cannot install anything yet.

automation_tools/__init__.py
```diff
@@ -1186,11 +1185,22 @@ def ak_install(admin_password=None, clean_beaker=True,
...
     print('Refreshing Subscription-manager.')
     run('subscription-manager refresh')
     print('Performing yum clean up.')
     run('yum clean all', quiet=True)
 
+    # Set yum stdout log level to be less verbose
+    set_yum_debug_level()
+
+    # Install some basic packages
+    update_basic_packages()
```
omaciel: The system should be fully registered by now and therefore you can install packages.

automation_tools/__init__.py
```diff
@@ -1434,14 +1444,20 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
-    execute(set_yum_debug_level, host=host)
+    # Create custom OS beaker repo
+    if os.environ.get('OS_UPGRADE_REPO'):
+        os_upgrade_repo = os.environ.get('OS_UPGRADE_REPO')
+        execute(create_custom_repos, rhel_candidate=os_upgrade_repo, host=host)
 
-    execute(install_prerequisites, host=host)
+    # If NOT using an activationkey, subscribe to CDN and install some
+    # basic required packages.
+    if distribution != 'satellite6-activationkey':
```
omaciel: When we're not using an `activationkey`, then we must register to the CDN and get some prerequisite packages from it.

automation_tools/__init__.py
```diff
@@ -1451,20 +1467,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
-        os_upgrade_repo = os.environ.get('OS_UPGRADE_REPO')
-        execute(create_custom_repos, rhel_candidate=os_upgrade_repo, host=host)
-    # Update the machine
-    execute(update_packages, host=host, warn_only=True)
 
     if distribution in ('satellite6-downstream', 'satellite6-iso'):
         execute(java_workaround, host=host)
 
-    # If using the internal dog food server's AK, unsubscribe from `CDN` first.
-    if distribution == 'satellite6-activationkey':
```
omaciel: We do not need to worry about unsubscribing since, if we are using an `activationkey`, then we won't subscribe anyhow.

omaciel:
@elyezer @kbidarkar @ehelms can you guys take a look here? The goal is to make sure that, if we're using an `activationkey` to install Satellite (such as when installing 6.3), then all yum calls should be done only after we're fully registered and subscribed to a host Satellite. We should **not** use CDN for these calls.

automation_tools/__init__.py
kbidarkar: update_basic_packages function takes care of it, hence the previous code is alright.

update_basic_packages, installs calls as below,

a) subscribe
b) install basic packages
c) unsubscribe

kbidarkar:
@omaciel , this PR is not required, the existing code is very much fine and should be taking care of things. Fix may be required at some place else, that is something which we need to figure out.

Currently for AK_Install, we are subscribing to CDN only to update the BASEOS packages, which is very much required as Satellite6.3 packages depends at times on the latest repos of BASEOS.

kbidarkar:
@omaciel, will investigate this issue and come up with a fix soon.

kbidarkar:
Sorry, I should explain this issue much clearer, I miss it at times.

This is a weird situation we are into with AK install procedure, 

1) If we decide we should be updating the image with latest BASEOS  packages, we may face this current issue. bind-utils package got updated to latest while bind package remained as it is.

2) Tomorrow if we do decide not to update the BASEOS and some package from Satellite6.3 repo, requires latest package as it's dependency from BASEOS, we may then require to again  change our code to update the packages.

Hence I was saying this change is not required.

I still feel it should be the DOGFOOD server which needs to be synced frequently and maintained upto date very often to avoid these issues. 

If there is no other solution, then we need this PR.

automation_tools/__init__.py
kbidarkar: @omaciel    :+1:   for moving it above

automation_tools/__init__.py
kbidarkar: Doing this will re-subscribe the box to CDN as it also calls in subscribe and unsubscribe within it.
You may want to drop this.

kbidarkar:
ACK, with pending comments.

To get past bind deps issue we need this change.

automation_tools/__init__.py
omaciel: But we don't want to subscribe to CDN...

automation_tools/__init__.py
omaciel: Don't we need to install these packages though?

omaciel:
@kbidarkar I replied to your comments with some questions of my own :)

automation_tools/__init__.py
kbidarkar: We are facing this as DOGFOOD is not 100% in sync with CDN.

Anyways, I was answering to "system is not yet registered" here. :)

automation_tools/__init__.py
kbidarkar: I looked at the `update_basic_package` method and it looks like we can add this line to get rid of the CDN subscription.
```
run('yum install -y yum-plugin-security yum-security', quiet=True)

```

elyezer: The subscribe should not be called inside `update_basic_packages`. It should be handled outside that task in order to make it atomic, all but the integration tasks (like product_install) must be atomic so the integration tasks can use them and orchestrate how the workflow is going to be.

kbidarkar: I agree with you @elyezer , but it is so, may be time to change all such instances.

kbidarkar: @omaciel, @elyezer, I volunteer to update the code and make it atomic, but for now let's remove this , replace the above run command and get this PR merged, will raise a PR to fix this throughout the codebase tomorrow. After the change I can make it use `update_basic_package` method, as then it will be atomic.

kbidarkar:
Closing this PR in favour of PR #437

PR #437, opened 2016-11-04
Fix cdn subs

kbidarkar:
This PR is same as PR https://github.com/SatelliteQE/automation-tools/pull/436 , but with a small change to update a method, so that PR 436 can be merged.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1173,7 +1173,6 @@ def ak_install(admin_password=None, clean_beaker=True,
 
     # Clean up and install with basic packages.
     clean_rhsm()
-    update_basic_packages()
 
     # Install the cert file
     run('yum -y localinstall {0}/pub/katello-ca-consumer-latest.noarch.rpm'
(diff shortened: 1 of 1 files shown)


lpramuk:
I dont get why for ak_install() these methods (install_prerequisites, update_packages...)
 must be incorporated in while for other *_install() are not? 

it can be simple and unified:

if AK subscribe to DOGFOOD
else subscribe to CDN

and then the rest of control flow is the same
- install_prerequisites
- set_yum_debug_level
- update_packages

it's getting messy here, and it doesnt have to

lpramuk:
But otherwise looks good. Still I can refactor it in successive PR :-)
ACK

automation_tools/__init__.py
```diff
@@ -1186,11 +1185,22 @@ def ak_install(admin_password=None, clean_beaker=True,
...
     )
     run(cmd)
+
+    # Refresh subscriptions and clean up YUM
     print('Refreshing Subscription-manager.')
     run('subscription-manager refresh')
     print('Performing yum clean up.')
     run('yum clean all', quiet=True)
 
+    # Set yum stdout log level to be less verbose
```
elyezer: These lines is being run on product install too, I would suggest to run the common things there and leave here only the tasks that are specific to the ak_install.

elyezer: Just saw @lpramuk comment. I think the changes could be request before merging, imo needing to rush is not an excuse to have code without quality but have a fast peace on updating the PR to get it merged.

PR #438, opened 2016-11-04
Call enable_satellite_repos only when distribution is not AK.

kbidarkar:


rplevka:
ACK pending `flake8` errors fixes

pondrejk:
ACK

PR #439, opened 2016-11-07
[Do Not Merge] Discovery packages have deps issue for sat6.3

kbidarkar:
a) Sat63 build use epel repos which brings in latest versioned rpm
b) Hence the rpm from sat63 get's ignored.
c) Also note that, syslinux-extlinux is required only for the epel
   rpm, which is actually not desired.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1552,12 +1552,16 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         # if we have ssh key to libvirt machine we can setup access to it
         if os.environ.get('LIBVIRT_KEY_URL') is not None:
             execute(setup_libvirt_key, host=host)
-        # setup_foreman_discovery
-        execute(
-            setup_foreman_discovery,
-            sat_version=satellite_version,
-            host=host
-        )
+        # Discovery packages being fetched for sat6.3 are currently latest
+        # from epel and not sattellite repos, hence broken.
+        # Temporary fix till we have this issue fixed.
+        if satellite_version != '6.3':
+            # setup_foreman_discovery
+            execute(
+                setup_foreman_discovery,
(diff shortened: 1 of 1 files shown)


kbidarkar:
This PR is raised to get some results rather than getting blocked on discovery setup issues.

Once discovery packaging/setup issues get fixed, we can revert this PR.

sghai:
ACK.. 

@omaciel This change is just to see automation results for sat6.3 on rhel6. Hope you'r fine with this change ?

automation_tools/__init__.py
```diff
@@ -1552,12 +1552,16 @@ def product_install(distribution, create_vm=False, certificate_url=None,
...
             execute(setup_libvirt_key, host=host)
-        # setup_foreman_discovery
-        execute(
-            setup_foreman_discovery,
-            sat_version=satellite_version,
-            host=host
-        )
+        # Discovery packages being fetched for sat6.3 are currently latest
+        # from epel and not sattellite repos, hence broken.
+        # Temporary fix till we have this issue fixed.
```
elyezer: Please have an issue to make sure we don't forget about it.

omaciel:
Guys, since 6.3 is at a crucial point right now, I'm going to vote against adding workarounds and instead requiring that we get a fix from our engineers. Also, as @elyezer pointed out, we should make sure to file issues whenever we do add a workaround so that we don't forget to revert changes.

sghai:
@omaciel : The real issue here is that we are using epel repo with Satellite builds. When we get standard satellite builds then this won't be an issue.

Also, currently, latest FDI which is going to ship with 6.3, is not build in brew so we are using old FDI rpm from 6.2. But in any case, neither the issue is with image nor with discovery plugin. Issue is with the repos that we are using.

Issue #440, opened 2016-11-08
repo_install() should enable rhel-6-server-optional-rpms


lpramuk:
Both {ak,repo}_install() should enable rhel-6-server-optional-rpms:
```
Error: Package: 1:livecd-tools-13.4.10-1.el6.x86_64 (Sat6-CI_CentOS_EPEL_6)
            Requires: syslinux-extlinux
  You could try using --skip-broken to work around the problem
  You could try running: rpm -Va --nofiles --nodigest
``` 
```syslinux-extlinux``` for rhel6 lives in ```optional``` repo

lpramuk:
ak_install() doesn't have to enable optional.
AK bears information of repos to enable (optioonal should be among them)

omaciel:
@lpramuk can this be closed since #441 was merged?

kbidarkar:
Yes, this can be closed now.

PR #441, opened 2016-11-08
Fix repo_install() to enable rhel-6-server-optional-rpms

lpramuk:
Fix repo_install() to enable rhel-6-server-optional-rpms

- fix missing dependecy for syslinux-extlinux on rhel6

(Issue #440)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1113,11 +1113,8 @@ def repofile_install(admin_password=None, run_katello_installer=True,
     run('wget -O /etc/yum.repos.d/satellite63.repo {0}'.format(repo_url))
 
     # Enable required repository
-    if os_version >= 7:
-        run(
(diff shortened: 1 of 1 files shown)


kbidarkar:
ACK

omaciel:
@lpramuk the subject of this PR implies that the `optional` repository should be enabled for RHEL 6 but it is not clear to me if we want to also enable it for RHEL 7. Your PR would always enable it, regardless of the OS version. Is that the desired outcome? If so, then ACK. Go ahead and merge it whoever gets to it first.

sghai:
@omaciel : We were enabling the optional repo for rhel7 but not for rhel6. However, as @lpramuk mentioned, we need to enable to the optional repo for rhel6 to resolve the dependency issue with rhel6 provisioning job. So we just removed the check.

sghai:
ACK

Issue #445, opened 2016-11-15
Unify / Simplify *_install() methods


lpramuk:
Please see my https://github.com/SatelliteQE/automation-tools/pull/437#issuecomment-258391015 for the idea
```
if distro=satellite6-activationkey:
    subscribe_dogfood()
else:
    subscribe()

<--- here we can continue unified  

install_prerequisites()
set_yum_debug_level()
update_packages()
```

lpramuk:
another reason is that:
os_upgrade_repo ```execute(create_custom_repos, rhel_candidate=os_upgrade_repo, host=host)``` needs to run after ```execute(install_prerequisites, host=host)```
(broken by PR https://github.com/SatelliteQE/automation-tools/pull/436)

elyezer:
#446 is merged. I think we can close this.

PR #446, opened 2016-11-16
Unify *_install() methods

lpramuk:
Unify *_install() methods
- unifies somewhat different ak_install() with other installs
- new subscribe_dogfood() used by ak_install()
- reorders various executes (silence yum, os update, os upgrade repo)

(Issue #445)


kbidarkar:
ACK, this change is required as we, no longer want ak_install method to update packages from CDN.

sghai:
ACK.. Looks good.

PR #448, opened 2016-11-18
Fixed the order of discovery setup task

sghai:
`setup_discovery_task` needs to be run at last otherwise, any other task like ostree which is re-running installer would re-set the discovery templates as well. A bz is already reported to fix this behavior.


abalakh:
Please add appropriate comment above `setup_discovery_task` so no one will accidentally move it again.

kbidarkar:
ACK

PR #455, opened 2016-11-21
Handle mounting of the right ISO file.

kbidarkar:
a) Some URL paths also have live iso files, which would cause
   issues, due to multiple iso file matches.
b) This fix should pick only a single ISO file.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1279,7 +1279,7 @@ def iso_install(
 
     # Create a 'check-out' folder, mount ISO to it...
     run('mkdir -p ~/ISO')
-    run('mount *.iso ~/ISO -t iso9660 -o loop')
+    run('mount *dvd1.iso ~/ISO -t iso9660 -o loop')
     # ...and run the installer script.
     with cd('~/ISO'):
         if check_gpg_signatures is True:


lpramuk:
@kedar Lets look at iso_download() to have constriction here (and not at mount time)

lpramuk:
Live Satellite ISOs? Can you give us an URL example why this is needed.

lpramuk:
iso_download() could return temp path of ISO to mount, then it would look like:
```
run('mount {0} ~/ISO -t iso9660 -o loop'
    .format(iso_download(iso_url))
)
```

lpramuk:
and if we still need constraint to *dvd1.iso then we better add it in iso_download() when matching pattern:
```
            if result.succeeded:
                # match either '<hash> *<iso_filename>'
                # or '{MD5|SHA1|SHA256} (<iso_filename>) = <hash>'
                iso_filename = search(
                    '\w+\s+[\*\(]?([^\s\)]+)', result).group(1)
                break
```

lpramuk:
To reviewers:
Please consider to merge https://github.com/SatelliteQE/automation-tools/pull/456 instead of this one.

kbidarkar:
@lpramuk, appears to have investigated more into this and had TESTED things in his PR, will close this.

lpramuk:
@kedar Thank you ;-)

PR #458, opened 2016-11-22
Pin rhel6 nightly version to 1.13 and use koji repos

lpramuk:
Devs are still building rhel6 bits, so lets continue to test these bits.


lpramuk:
It's a still mess even with Copr Qpid
```
Error: Package: foreman-proxy-1.13.0-1.el6.noarch (foreman_koji)
           Requires: rubygem(sinatra)
```
```rubygem-sinatra``` is only for rhel7 ?

PR #405, opened 2016-09-08
Attemp to mitigate "Error: INVALID_ZONE"

lpramuk:
Right after starting firewalld service we are submitting firewall-cmd commands.
Sometimes they fail sometimes not.
```
[hostname] run: systemctl start firewalld
[hostname] run: for port in $(firewall-cmd --permanent --list-ports); do firewall-cmd --permanent --remove-port="$port"; done
[hostname] out: Error: INVALID_ZONE
[hostname] out: Error: INVALID_ZONE
[hostname] out: 

Fatal error: run() received nonzero return code 112 while executing!
```
I added in between some checks for zones and some sleep. Sleep fixes things. Magic :-)

PR #473, opened 2017-01-03
Attemp #2 to mitigate "Error: INVALID_ZONE"

lpramuk:
Attemp No.2 to mitigate "Error: INVALID_ZONE"
- check for permanent zones (--permanent) since we check/modify the permanent ones
- restart firewalld service and wait (mere waiting doesn't help)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -449,8 +449,11 @@ def setup_firewall(definitions=None, flush=True):
             run('systemctl start firewalld')
         # FIXME attempt to mitigate "Error: INVALID_ZONE"
         for i in range(5):
-            if run('firewall-cmd --list-all-zones', warn_only=True).succeeded:
+            if run(
+                'firewall-cmd --permanent --list-all-zones', warn_only=True
+            ).succeeded:
                 break
+            run('systemctl restart firewalld')
             time.sleep((i+1) * 5)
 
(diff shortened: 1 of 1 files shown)


lpramuk:
attempt #1 was merged as PR #405

omaciel:
@kbidarkar mind reviewing this when you have a chance?

lpramuk:
I hit it today, so it forced me to test it and I happened to resolved it by these changes.

At last we need to get rid of katellovirt module which is causing this (interaction with firewalld).
We better install libvirt by set of cmds that we control over easily.

kbidarkar:
@lpramuk, I just checked the jobs to find the issue where we could see this with logs, but didn't find one. While using which job did you face this issue? It's seems to be working fine currently for the jobs.

kbidarkar:
But, yeah we would need to use the "permanent" option to firewall-cmd as we are updating the firewall ports for persist-ency across reboot.

lpramuk:
this issue arises if you are reusing machine or your instance is not clean OS install.
Mainly in satellite6 installer (not in automation)

kbidarkar:
I think, currently we do not provide support for reusing a machine with jenkins jobs, without re-installing the OS.
There was no such effort since the beginning and we need to work towards it if we need to provide support for it to happen.

Anyways, we require this, as this would correct the current commands being used.

ACK

PR #475, opened 2017-01-05
Don't override INTERNAL for 6.3

lpramuk:
Allow to use ```INTERNAL``` for 6.3 as it is valid and **working** option. 

Even with this change, ```INTERNAL AK``` will still be the default as it is listed in all jenkins choices as the first option.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1400,16 +1400,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     }
     distribution = distribution.lower()
 
-    # Make sure downstream_install only is called for sat6.1 and sat6.2, as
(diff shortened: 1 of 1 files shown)


kbidarkar:
Thought sat63 was just AK and repo_file install , so had disabled it.  If needed and on popular demand we can enable it back.

ACK

omaciel:
@lpramuk @kbidarkar can you guys check with @ehelms to make sure that internal repos would work for 6.3 too?

lpramuk:
@omaciel it works due to my MR#20 and I tested it, in my branch.

INTERNAL is one repo method
INTERNAL REPO was mutiple repos method, now turned into one repo method, so basically is now the same as the one above

san7ket:
ACK

PR #484, opened 2017-01-11
Nightly has --capsule-puppet renamed to --foreman-proxy-puppet

lpramuk:
Transition to foreman-proxy named options was done gradually,
thus we need more proxy* vars and more ifs setting them appropriately

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -338,6 +338,10 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         proxy = 'capsule'
     else:
         proxy = 'foreman-proxy'
+    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1', '6.2', '6.3'):
+        proxypuppet = 'capsule'
+    else:
+        proxypuppet = 'foreman-proxy'
 
     installer_options = {
         '{0}-dns'.format(proxy): 'true',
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -338,6 +338,10 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         proxy = 'capsule'
     else:
         proxy = 'foreman-proxy'
+    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1', '6.2', '6.3'):
```
omaciel: Are you sure that the command changed for all versions of Satellite? I can attest that Satellite 6.2.7 shows several entries for `--foreman-proxy'

lpramuk: Think opposite way here:
the option stayed same for all these versions and changed for nightly which is not listed 
(nightly,6.4, 6.5)

(I know that due to rebase to 1.15 probably the value '6.3' would need to be removed later on)

omaciel: Thanks @lpramuk ACK

kbidarkar:
Suggesting we move this under katello_installer, https://github.com/lpramuk/automation-tools/blob/5f01b5c5351dc68ac6cd2e8a7fa52579b05e1f37/automation_tools/__init__.py#L2157

lpramuk:
@kbidarkar  I dont get your point? I'm not refactoring I'm just updating one option the same way as it is already implemented

kbidarkar:
Hey, I am talking of the Lines 341 to 344, the place where `proxypuppet` variable is defined, I am suggesting just move this part to katello_installer function.

kbidarkar:
ACK

PR #485, opened 2017-01-11
Fix reboot in Personal Satellite zStream upgrade

jyejare:
This PR fix the personal Satellite Upgrade for zStream builds.

Fix:
Ideally we dont reboot personal satellite/capsules during upgrade due to its unknown environment and configuration due to which we cant provide timeout for reboot. So currently only our RHEVM satellite/capsule are eligible for reboots.


sghai:
ACK

PR #489, opened 2017-01-12
Collect Foreman Debug on Upgrade Fail

jyejare:
This PR helps collect Foreman Debug even the Upgrade Job failed with an exception during Satellite and Capsule Upgrade.

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -166,31 +166,44 @@ def product_upgrade(product):
         to_version = os.environ.get('TO_VERSION')
         sat_host, cap_hosts, clients6, clients7 = setup_products_for_upgrade(
             product, os.environ.get('OS'))
-        with LogAnalyzer(sat_host):
-            if from_version != to_version:
-                current_version = execute(
-                    get_sat_version, host=sat_host)
-                if not current_version == to_version:
-                    execute(satellite6_upgrade, host=sat_host)
-            elif from_version == to_version:
-                execute(satellite6_zstream_upgrade, host=sat_host)
-            # Generate foreman debug on satellite after upgrade
+        try:
+            with LogAnalyzer(sat_host):
+                if from_version != to_version:
+                    current_version = execute(
+                        get_sat_version, host=sat_host)
+                    if not current_version == to_version:
+                        execute(satellite6_upgrade, host=sat_host)
+                else:
+                    execute(satellite6_zstream_upgrade, host=sat_host)
+                # Generate foreman debug on satellite after upgrade
+                execute(foreman_debug, 'satellite_{}'.format(sat_host),
+                        host=sat_host)
+                if product == 'capsule':
+                    for cap_host in cap_hosts:
+                        try:
+                            with LogAnalyzer(cap_host):
+                                if from_version != to_version:
+                                    execute(satellite6_capsule_upgrade,
+                                            cap_host, host=cap_host)
+                                elif from_version == to_version:
+                                    execute(satellite6_capsule_zstream_upgrade,
+                                            host=cap_host)
+                                # Generate foreman debug on capsule postupgrade
+                                execute(
+                                    foreman_debug,
+                                    'capsule_{}'.format(cap_host),
+                                    host=cap_host)
+                        except Exception:
+                            # Generate foreman debug on failed capsule upgrade
+                            execute(
+                                foreman_debug, 'capsule_{}'.format(cap_host),
+                                host=cap_host)
+                            raise
+                if product == 'client':
+                    satellite6_client_upgrade('rhel6', clients6)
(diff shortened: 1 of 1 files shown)


jyejare:
@SatelliteQE/quality-engineers , The PR is up for review ! Please visit !!

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -166,31 +166,44 @@ def product_upgrade(product):
...
+                                execute(
+                                    foreman_debug,
+                                    'capsule_{}'.format(cap_host),
+                                    host=cap_host)
+                        except Exception as caperr:
+                            # Generate foreman debug on failed capsule upgrade
+                            execute(
+                                foreman_debug, 'capsule_{}'.format(cap_host),
+                                host=cap_host)
+                            raise Exception(caperr)
```
elyezer: You can just call `raise` here.

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -166,31 +166,44 @@ def product_upgrade(product):
...
-                                    host=cap_host)
-                        # Generate foreman debug on capsule after upgrade
-                        execute(
-                            foreman_debug,
-                            'capsule_{}'.format(cap_host),
-                            host=cap_host)
-            if product == 'client':
-                satellite6_client_upgrade('rhel6', clients6)
-                satellite6_client_upgrade('rhel7', clients7)
+            raise Exception(saterr)
```
elyezer: Just call `raise`

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -166,31 +166,44 @@ def product_upgrade(product):
...
+                    if not current_version == to_version:
+                        execute(satellite6_upgrade, host=sat_host)
+                elif from_version == to_version:
+                    execute(satellite6_zstream_upgrade, host=sat_host)
+                # Generate foreman debug on satellite after upgrade
+                execute(foreman_debug, 'satellite_{}'.format(sat_host),
+                        host=sat_host)
+                if product == 'capsule':
+                    for cap_host in cap_hosts:
+                        try:
```
elyezer: This try/except block is not needed since there is another one already wrapping and that will be able to capture the foreman debug.

elyezer: Having this nested try/except will actually get foreman-debug two times:

```python
In [1]: try:
   ...:     try:
   ...:         raise Exception()
   ...:     except Exception:
   ...:         print('inner')
   ...:         raise
   ...: except Exception:
   ...:     print('outer')
   ...:     raise
   ...: 
inner
outer
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-1-84e98f4aa4f9> in <module>()
      1 try:
      2     try:
----> 3         raise Exception()
      4     except Exception:
      5         print('inner')

Exception: 

```

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -166,31 +166,44 @@ def product_upgrade(product):
...
-                execute(satellite6_zstream_upgrade, host=sat_host)
-            # Generate foreman debug on satellite after upgrade
+        try:
+            with LogAnalyzer(sat_host):
+                if from_version != to_version:
+                    current_version = execute(
+                        get_sat_version, host=sat_host)
+                    if not current_version == to_version:
+                        execute(satellite6_upgrade, host=sat_host)
+                elif from_version == to_version:
```
elyezer: This can safely be a `else` block since the first condition is that `from_version` != `to_version`

automation_tools/satellite6/upgrade/__init__.py
jyejare: The inside try .. except is to capture the foreman-debug logs on capsule(s) and only if capsule upgrade is opted.

I cannot collect foremen debug logs outside the capsules loop onto which the capsule upgrade is running in loops. And I a collecting foreman-debug on each capsule in the loop.

jyejare:
@elyezer , I worked on your comments. Please revisit to re-review.

automation_tools/satellite6/upgrade/__init__.py
elyezer: So you should not re-raise the exception otherwise it will stop the for loop and will be handled by the outer try/except block.

jyejare: @elyezer , Actually That is the main intention, I wanted to capture the foreman-debug for both satellite, capsule  as well.

OR else, may be I am not getting what you are saying :)

elyezer: Now I got the idea, I saw they are capturing for different hosts. Thanks for the clarification and sorry about this back and forth.

PR #492, opened 2017-01-19
Start docker.service with warn_only=True

lpramuk:
Start docker.service with warn_only=True
Docker 1.12 service fails to start, as a result the whole provisioning/installation job fails.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -280,7 +280,8 @@ def setup_default_docker():
 
     # enable the service as it is disabled by default on RHEL7
     manage_daemon('enable', 'docker', pty=(os_version >= 7))
-    manage_daemon('restart', 'docker', pty=(os_version >= 7))
+    manage_daemon('restart', 'docker', pty=(os_version >= 7),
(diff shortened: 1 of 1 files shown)


lpramuk:
The root cause https://github.com/projectatomic/container-selinux/pull/26

kbidarkar:
@lpramuk, feel free to remove the "Do Not Merge", once done.

lpramuk:
@kbidarkar ```warn_only=True``` is now conditional as we have discussed.

elyezer:
So when the docker is not started then some docker related tests is going to fail, are we ok with that?

I have no personal opinion here just wanted to make sure we are aware of that.

lpramuk:
@elyezer  Probably we are ok.

I'm guessing if I tried to workaround it (I have already solution) that PR would get NACK

elyezer:
@lpramuk if there is a BZ associated maybe having the workaround while that BZ is open. I just wanted to raise a possible issue of having some failures related to the docker not being installed.

lpramuk:
@elyezer I agree and it's done this way already - associated with BZ 1414821

PR #494, opened 2017-01-24
Fix the capital case issue with hostname

### LINKED ISSUE #493: Fix attaching custom product subscription to Docker Host
We are facing issue in :
Attaching custom product subscription to the host created on Docker. Following error has been thrown while attaching subscription to the docker host;
Error: Host not found

jyejare:
We faced an issue while attaching custom product of latest tools repo(for upgrade) to the host created on docker. Because host was not identified to run hammer attach command.

Basically It was a capital case issue in docker hostname as below:
We were passing hostname as '0DockerClientrhel7(Please not D and C are capital here)' but the satellite was converting this into 0dockerclientrhel7(Note that D and C is now lower. And we were finding or attaching the subscriptions to the given hostname which is '0DockerClientrhel7' which doesnt exists.

So fixed the same by lowering the hostname while giving itself.

Fixes #493


sghai:
ack

kbidarkar:
ACK

PR #501, opened 2017-02-01
Client Registeration race condition fix with time sleeps

jyejare:
This will fix the race condition issue happening during registering many clients at a time.


jyejare:
@rplevka , Even though there is bug regarding generating multiple clients at a time, but this is not a real time scenario with customers. And so keeping sleeps will make it at most equal to customer scenarios and even 5 seconds of sleep is much smaller.

rplevka:
@jyejare i'm not asking you to remove the `sleep`, just asking you to decorate that workaround properly.

sghai:
ACK

rplevka:
ACK

PR #504, opened 2017-02-10
Changes to support another rhevm3 configs

sghai:
- Changes to use `create_rhevm_instance` fn w/ another rhevm having diff cluster name.
- Fetch FQDN of instance if cluster=`SAT-QE`. We can fetch the fqdn only if rhev-agent is installed on instances. For now, all instances created under `SAT-QE` cluster will have the agent installed so the change only applicable for  `SAT-QE` cluster.


kbidarkar:
ACK

PR #507, opened 2017-02-13
Exporting instance name through file

sghai:


sghai:
@elyezer @kbidarkar : guys.. please review the PR and suggest if there is any other way to export the `vm_fqdn` variable so that we can use the value of it w/ `robottelo-CI`.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -300,7 +300,12 @@ def create_rhevm_instance(instance_name, template_name, datacenter='Default',
                 vm_fqdn = rhevm_client.vms.get(
                     name=instance_name).get_guest_info().get_fqdn()
                 print('\t Instance FQDN : %s' % (vm_fqdn))
-                return vm_fqdn
+                # We need value of vm_fqdn so that we can use it with CI
+                # For now, we are exporting it as a variable value
+                # and source it to use via shell script
+                file_path = "/tmp/rhev_instance.txt"
+                with open(file_path, 'w') as file:
+                    file.write('export SAT_INSTANCE_FQDN={0}'.format(vm_fqdn))
```
sghai: source the file here in CI like: https://github.com/SatelliteQE/robottelo-ci/pull/489/commits/fa591019cd7a992f70731d53cce29bf693230501#diff-ac1c90cc2e4c0f362d2665a9877681f3R45

elyezer:
I don't have any other suggestion, the code looks fine to me.

sghai:
thanks @elyezer 

@jyejare @kbidarkar : please review guys.

kbidarkar:
ACK

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -300,7 +300,12 @@ def create_rhevm_instance(instance_name, template_name, datacenter='Default',
                 vm_fqdn = rhevm_client.vms.get(
                     name=instance_name).get_guest_info().get_fqdn()
                 print('\t Instance FQDN : %s' % (vm_fqdn))
-                return vm_fqdn
```
kbidarkar: I think, this could be simplified by using the below

`return run(export SAT_INSTANCE_FQDN={0}'.format(vm_fqdn))`

kbidarkar: Another benefit could be that this way, we may not require to source the file at robotello-ci end.

If possible, can you please try this out?

kbidarkar: But may be with this change, it may not help with readability of code at the robotello-ci end.

sghai: Thanks @kbidarkar.  Nop.. that won't work because python exports the variable in one shell and when bash tries to read the var, it won't get that as it runs the script in another shell.

PR #516, opened 2017-02-22
Added support to configure named service & finish template for osp

kbidarkar:
a) Would configure the forward and reverse zones needed for
   provisioning hosts on OpenStackPlatform.
b) Will clone and update the finish template with reading of
   sshkeys required for accessing the provisioned hosts.


sghai:
ACK

san7ket:
ACK

PR #520, opened 2017-02-24
Fix wrong message of Capsule Upgrade Highlight from 6.1 to 6.2

### LINKED ISSUE #519: Fix: The false Highligh log of Capsule is Not upgraded
There is a false Highlight log regarding Capsule Upgrade, when upgrading from 6.1 to 6.2.
Fix the issue on top priority.

jyejare:
This PR:

1. Wrong message for Capsule Upgrade appeared for 6.1 to 6.2 upgrade, saying the capsule is not upgraded but the capsule is actually upgraded. Closes #519 
2. Modified 'CLIENT UPGRADE: <rhel_version>' to '<rhel_version> CLIENTS Upgrade' for better meaning of rhel7 and rhel6 clients.
3. get_Sat_cap_ver function modifeid to return None with reason if no version found.

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -179,14 +179,14 @@ def product_upgrade(product):
             product, os.environ.get('OS'))
         try:
             with LogAnalyzer(sat_host):
-                current = str(execute(
-                    get_sat_cap_version, 'sat', host=sat_host)[sat_host])
+                current = execute(
+                    get_sat_cap_version, 'sat', host=sat_host)[sat_host]
                 if from_version != to_version:
                     execute(satellite6_upgrade, host=sat_host)
                 else:
                     execute(satellite6_zstream_upgrade, host=sat_host)
-                upgraded = str(execute(
-                    get_sat_cap_version, 'sat', host=sat_host)[sat_host])
+                upgraded = execute(
+                    get_sat_cap_version, 'sat', host=sat_host)[sat_host]
                 if LooseVersion(upgraded) > LooseVersion(current):
                     logger.highlight(
                         'The Satellite is upgraded from {0} to {1}'.format(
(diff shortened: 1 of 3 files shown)


automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -214,7 +214,7 @@ def product_upgrade(product):
                                 upgraded = str(execute(
                                     get_sat_cap_version, 'cap', host=cap_host
                                     )[cap_host])
-                                if current:
+                                if current != 'None':
```
kbidarkar: I feel if 'None' is a string then something is wrong here.
I feel we shouldn't be checking for `current != 'None'` this.

jyejare: actually the current value will be passed to LooseVersion function to compare previuos version and upgraded version. So while fetching the version I get it string type(as https://github.com/jyejare/automation-tools/blob/dcfd2f1f84a6d3472ff77be58fade467b82f20a7/automation_tools/satellite6/upgrade/__init__.py#L182). 
So that I need not to type cast it everytime(e.g https://github.com/jyejare/automation-tools/blob/dcfd2f1f84a6d3472ff77be58fade467b82f20a7/automation_tools/satellite6/upgrade/__init__.py#L190)

kbidarkar: AFAIK, we should not do this type cast `None` to `'None'` string and then check for if `current != 'None'`. This sounds like something is not right.
I would suggest we typecast it here , https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/satellite6/upgrade/tasks.py#L721 
and 
also let that function return a `None` value if version is `NULL` or `''` or not found or 'Unavailable' for some reason. For example if `satellite` package itself is missing for some reason.

```
if version != 'Unavailable':
    return str(version)
```
Can you please check if we can do this?  
So with this change if `version == ''` or `version == 'Unavailable'` , `get_sat_cap_version` will return `None`. So that we should be able to do just `if current:`.

kbidarkar: @jyejare, Also do you think we may want to check for `version == ''` and return `None` ?

elyezer: I think the `get_sat_cap_version` should raise an exception when the version can't be figured out since that knowing the version is required. Doing that cast, for None as string seems a bad code smell to me.

jyejare: @elyezer , We dont want to rasie an exception, giving an example on Capsule 6.1 and less versions it is just imposssible to get version. But in 6.2 abd above both sat and cap versions are accessible.

jyejare:
@kbidarkar @elyezer Hey I updated the get_sat_cap_version function accordingly. Please review and sorry for bad code practiced due to hurry to fix that asap.

Issue #526, opened 2017-03-02
Fix yum annoying logs in client upgrade


jyejare:
Client Upgrade failing with following annoying logs in docker clients:

[10.8.104.58] run: docker exec 3af8e26c8f7145181c32bb369f5bff0eff6922c3eb8fbd680706038a87b0e315 yum clean all
[10.8.104.58] out: Loaded plugins: product-id, subscription-manager
[10.8.104.58] out: Existing lock /var/run/yum.pid: another copy is running as pid 72.
[10.8.104.58] out: Another app is currently holding the yum lock; waiting for it to exit...
[10.8.104.58] out:   The other application is: yum
[10.8.104.58] out:     Memory :  48 M RSS (354 MB VSZ)
[10.8.104.58] out:     Started: Wed Mar  1 10:41:04 2017 - 00:35 ago
[10.8.104.58] out:     State  : Uninterruptible, pid: 72
[10.8.104.58] out: Another app is currently holding the yum lock; waiting for it to exit...
[10.8.104.58] out:   The other application is: yum
[10.8.104.58] out:     Memory :  48 M RSS (354 MB VSZ)
[10.8.104.58] out:     Started: Wed Mar  1 10:41:04 2017 - 00:37 ago
[10.8.104.58] out:     State  : Uninterruptible, pid: 72
[10.8.104.58] out: Another app is currently holding the yum lock; waiting for it to exit...

jyejare:
The referenced PR has been merged so closing this issue.

PR #528, opened 2017-03-02
Fix annoying yum lock already taken issue in client upgrade

jyejare:
Fix issue mentioned in subject by increasing the sleep time by the time yum locks on all clients will be free.

We will see how this goes, if this doesnt works we may have to wait until yum unlocks on each client or kill yum process.

### DIFF
--- a/automation_tools/satellite6/upgrade/client.py
+++ b/automation_tools/satellite6/upgrade/client.py
@@ -174,12 +174,12 @@ def satellite6_client_setup():
                 host=sat_host
             )
         # Refresh subscriptions on clients
-        time.sleep(5)
+        time.sleep(30)
         execute(
             refresh_subscriptions_on_docker_clients,
             clients6.values(),
             host=docker_vm)
-        time.sleep(5)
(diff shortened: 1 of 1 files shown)


jyejare:
closes #526

automation_tools/satellite6/upgrade/client.py
```diff
@@ -174,12 +174,12 @@ def satellite6_client_setup():
                 host=sat_host
             )
         # Refresh subscriptions on clients
-        time.sleep(5)
+        time.sleep(30)
```
sghai: I understood the issue w/ yum lock. However do we really need 30 sec sleep ? Just asking if lowering down to 15secs would work ?

jyejare: I think its ok, its just a one time sleep just before 'yum clean all' on all the clients. However this is not something 'number of clients' * 30. Its just one single 30 seconds.

sghai:
ACK..

kbidarkar:
ACK

Issue #513, opened 2017-02-20
Upgrade tasks - needs to be performed w/ rhev templates


sghai:
- [x] Images needs to be updated so that all manifest tests/RHEL repo enabling tests should succeed

- [x] Run this task once `setup_alternate_capsule_ports`

- [x] Create rhev_template task so that we can create templates via rhev api's

sghai:
First two tasks required to improve the tier1 test results post upgrade.

sghai:
Third one related to `rhev_template` is required to re-create the template everytime a new version is released. We may add this as conditional check w/ `satellite6_upgrader` job. If the create_template check is enabled, the job itself will do a sanity check post-upgrade and creates the templates w/ newly released version.

Please feel free to either create new tasks or consider this task as a tracker.

sghai:
For first task, this needs to be run on upgrade instances:  https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L399

san7ket:
The tasks for the creating rhev template, is merged. But blocked on creating a job for this , unless we get a mechanism to handle the paused and errored tasks. We can still use "fab" to run the task, So we could say we have a partial automation around this.

san7ket:
Closing this ,as the PR is merged

PR #533, opened 2017-03-06
Tasks for checking the upgarde boxes and creating templates

san7ket:
Created tasks, for doing a sanity check on the upgrade boxes and create their templates after upgrading them to next z-stream .
one of the step for issue https://github.com/SatelliteQE/automation-tools/issues/513

### DIFF
--- a/automation_tools/satellite6/upgrade/template.py
+++ b/automation_tools/satellite6/upgrade/template.py
@@ -1,0 +1,121 @@
+from automation_tools.satellite6.hammer import hammer, set_hammer_config
+from automation_tools.satellite6.upgrade import \
+    check_necessary_env_variables_for_upgrade
+from automation_tools.satellite6.upgrade.tasks import \
+    get_rhevm_client, wait_till_rhev_instance_status
+from fabric.api import run, execute
+from ovirtsdk.xml import params
+from time import sleep
+import os
+import sys
+
+
+# Restart the Services
+def service_restart():
+    try:
+        services = run('katello-service restart')
+        if services.return_code > 0:
+            sys.exit(1)
+    except Exception as ex:
+        print ex
+        sys.exit(1)
+
+
+# Running capsule sync on external capsule
+def check_capsule():
+    set_hammer_config(user=None, password=None)
+    try:
+        capsules = hammer('capsule list')
+        cap_id = 0
+        for item in capsules:
+            if str(item[u'name']).__contains__("capsule"):
+                cap_id = str(item[u'id'])
+                print cap_id
+        check = hammer("capsule refresh-features --id {0}".format(cap_id))
+        print check[u'message']
+        if check.return_code == 0:
+            print "running capsule sync"
+    except Exception as ex:
+        print ex
+        sys.exit(1)
+
+
+# Check if ntpd is running
+def check_ntpd():
+    try:
+        ntpd_check = run("service ntpd status", warn_only=True)
+        if ntpd_check.return_code > 0:
+            run("service ntpd start")
+            run("chkconfig ntpd on")
+    except Exception as ex:
+        print ex
+        sys.exit(1)
+
+
+# Check for Vm's and create their templates
+def create_template(host, cluster, new_template, storage):
+    """ Create template from vmName """
+    try:
+        get_client = get_rhevm_client()
+        storage_domain = get_client.storagedomains.get(name=storage)
+        size = storage_domain.get_available() / 1024 / 1024 / 1024
+        vm = get_rhevm_client().vms.get(host)
+        if size > 300 and vm:
+            try:
+                vm.stop()
+                wait_till_rhev_instance_status(host, 'down', 5)
+                print 'Waiting for VM to reach Down status'
+                get_client.templates.add(
+                    params.Template(name=new_template,
+                                    vm=get_client.vms.get(host),
+                                    cluster=get_client.clusters.get(cluster)))
+                while get_client.vms.get(host).status.state != 'down':
+                    sleep(1)
+                    "Template creation in Progress"
+            except Exception as ex:
+                get_client.disconnect()
+                print 'Failed to Create Template from VM:\n%s' % str(ex)
+        else:
+            get_client.disconnect()
+            print "Low Storage cannot proceed or VM not found"
+            sys.exit(1)
+    except Exception as ex:
+        get_client.disconnect()
+        print ex
+        sys.exit(1)
+
+
+# Fabric task
+def validate_and_create():
+    sat_host = os.environ.get('RHEV_SAT_HOST')
+    cap_host = os.environ.get('RHEV_CAP_HOST')
+    cluster = 'Default'
+    storage = os.environ.get('RHEV_STORAGE')
+    for host in sat_host, cap_host:
+        if host == sat_host:
+            image = os.environ.get('RHEV_SAT_IMAGE')
+            sat_host = os.environ.get('RHEV_SAT_HOST')
+            new_template = image + "_new"
+            if check_necessary_env_variables_for_upgrade('capsule'):
+                try:
+                    execute(check_ntpd, host)
+                    execute(service_restart, host)
+                    execute(check_capsule, host)
+                    create_template(sat_host, cluster, new_template, storage)
+                except Exception as ex:
+                    print ex
+                    print "Satellite Validation or Template creation Failed"
+                    sys.exit(1)
+        else:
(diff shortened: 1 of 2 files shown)


fabfile.py
```diff
@@ -100,3 +100,6 @@
     host_pings,
     reboot
 )
+from automation_tools.satellite6.upgrade.template import (
+    validate_and_create
+)
```
elyezer: Add a new line here.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
+from automation_tools.satellite6.hammer import hammer, set_hammer_config
```
elyezer: Would be great to have a docstring explaining what code will be in this module.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+from automation_tools.satellite6.upgrade import \
```
elyezer: Avoid using `\`, use `(` and `)` like:

```python
from some_module import (
    one_thing_to_import,
)

File: automation_tools/satellite6/upgrade/template.py [hunk 4]
```diff
@@ -0,0 +1,121 @@
...
+    check_necessary_env_variables_for_upgrade
+from automation_tools.satellite6.upgrade.tasks import \
+    get_rhevm_client, wait_till_rhev_instance_status
+from fabric.api import run, execute
+from ovirtsdk.xml import params
+from time import sleep
+import os
```
elyezer: `import` statements should come before any `from ... import ...```

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+import sys
+
+
+# Restart the Services
+def service_restart():
```
elyezer: Please add docstring

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+    try:
+        services = run('katello-service restart')
+        if services.return_code > 0:
+            sys.exit(1)
```
elyezer: Better to print something here saying to the user why the code have exited.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+    except Exception as ex:
```
elyezer: The try/except block is not needed. If the exception is not catch then it will sys.exit and will print the exception information.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+        print ex
+        sys.exit(1)
+
+
+# Running capsule sync on external capsule
+def check_capsule():
```
elyezer: Please add docstring.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+        cap_id = 0
+        for item in capsules:
+            if str(item[u'name']).__contains__("capsule"):
+                cap_id = str(item[u'id'])
+                print cap_id
+        check = hammer("capsule refresh-features --id {0}".format(cap_id))
+        print check[u'message']
+        if check.return_code == 0:
+            print "running capsule sync"
+    except Exception as ex:
```
elyezer: Same here, no need to wrap in try/except

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+        print ex
+        sys.exit(1)
+
+
+# Check if ntpd is running
+def check_ntpd():
```
elyezer: Please add docstring.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+    try:
+        ntpd_check = run("service ntpd status", warn_only=True)
+        if ntpd_check.return_code > 0:
+            run("service ntpd start")
+            run("chkconfig ntpd on")
+    except Exception as ex:
```
elyezer: No need to have a try/except block

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+                get_client.templates.add(
+                    params.Template(name=new_template,
+                                    vm=get_client.vms.get(host),
+                                    cluster=get_client.clusters.get(cluster)))
+                while get_client.vms.get(host).status.state != 'down':
+                    sleep(1)
+                    "Template creation in Progress"
+            except Exception as ex:
+                get_client.disconnect()
+                print 'Failed to Create Template from VM:\n%s' % str(ex)
```
elyezer: Shouldn't the code stop if the template has failed to be created?

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+        if ntpd_check.return_code > 0:
+            run("service ntpd start")
+            run("chkconfig ntpd on")
+    except Exception as ex:
+        print ex
+        sys.exit(1)
+
+
+# Check for Vm's and create their templates
+def create_template(host, cluster, new_template, storage):
```
elyezer: This function is a little bit confusing, could you explain what you are trying to accomplish?

Having those two try/except block seems to be redundant

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+            print "Low Storage cannot proceed or VM not found"
+            sys.exit(1)
+    except Exception as ex:
+        get_client.disconnect()
+        print ex
+        sys.exit(1)
+
+
+# Fabric task
+def validate_and_create():
```
elyezer: Please add docstring

automation_tools/satellite6/upgrade/template.py
elyezer: The comment above the function could be used as the docstring here and on the other functions.

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+    sat_host = os.environ.get('RHEV_SAT_HOST')
+    cap_host = os.environ.get('RHEV_CAP_HOST')
```
elyezer: Is is ok if those env vars are not defined? I think the code should check and fail if they are not provided since the remaining of the code is relying on those being defined.

automation_tools/satellite6/upgrade/template.py
jyejare: Rename it to 'What to validate and create' ?

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+    cluster = 'Default'
+    storage = os.environ.get('RHEV_STORAGE')
+    for host in sat_host, cap_host:
+        if host == sat_host:
+            image = os.environ.get('RHEV_SAT_IMAGE')
+            sat_host = os.environ.get('RHEV_SAT_HOST')
+            new_template = image + "_new"
+            if check_necessary_env_variables_for_upgrade('capsule'):
```
jyejare: Why are you checking necessary variables for capsule as you are operating with satellite ?

automation_tools/satellite6/upgrade/template.py
```diff
@@ -0,0 +1,121 @@
...
+        sys.exit(1)
+
+
+# Fabric task
+def validate_and_create():
+    sat_host = os.environ.get('RHEV_SAT_HOST')
+    cap_host = os.environ.get('RHEV_CAP_HOST')
+    cluster = 'Default'
+    storage = os.environ.get('RHEV_STORAGE')
+    for host in sat_host, cap_host:
```
jyejare: I believe you dont need for loop here, simply if ..else is enough.

san7ket:
Will soon address the comments, but this also requires handling and cleaning the errored tasks , before creating templates , the mechanism for this is to be delivered by foreman-maintain project

san7ket:
Will address the comments onto new PR in sat-upgrade project. Will raise this PR in sat-upgrade project as there's refactoring of structure on new project.

san7ket:
The new PR is raised in satellite6-upgrade project. Please revisit https://github.com/SatelliteQE/satellite6-upgrade/pull/16

PR #535, opened 2017-03-08
dns-id is not associated at this stage.

kbidarkar:
This is not required in automation-tools as this exact task is performed much later using the satellite6-populate job, https://github.com/SatelliteQE/robottelo-ci/blob/master/scripts/satellite6-populate-template.sh#L230


sthirugn:
ACK

san7ket:
ACK

PR #537, opened 2017-03-14
IDM and AD EXTERNAL_AUTH are now mutually exclusive

kbidarkar:
Earlier IDM_EXTERNAL_AUTH and AD_EXTERNAL_AUTH were boolean values
and both could be selected. Where as only one can be selected as
they are mutually exclusive and the only way to get EXTERNAL_AUTH
work with both of them is by establishing CROSS-REALM TRUST
between them.


sthirugn:
ACK

PR #542, opened 2017-03-23
Fix yum lock non available issue for client subscriptions

jyejare:
Sleep time is added to wait untill the docker client generation tasks frees up the yum lock for further tasks.

### DIFF
--- a/automation_tools/satellite6/upgrade/client.py
+++ b/automation_tools/satellite6/upgrade/client.py
@@ -141,12 +141,16 @@ def satellite6_client_setup():
             int(clients_count)/2,
             host=docker_vm
         )[docker_vm]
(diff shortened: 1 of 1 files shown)


san7ket:
ACK

automation_tools/satellite6/upgrade/client.py
```diff
@@ -141,12 +141,16 @@ def satellite6_client_setup():
             int(clients_count)/2,
             host=docker_vm
         )[docker_vm]
+        # Adding sleep to freed up yum lock taken by above client generate task
+        time.sleep(30)
```
sghai: Let's see how it goes w/ sleep otherwise we need to revisit again for more suitable fix..

sghai:
ACK

Issue #546, opened 2017-03-24
Fix Client Upgrade highlight log


jyejare:
Fix client upgrade highlight log that shows no change in katello-agent package version post upgrade.

jyejare:
Closing this issue as ref. PR is merged.

PR #557, opened 2017-04-04
Clients from_version calculation is removed in upgrade logging

jyejare:
Removed calculation of clients from_version or previous version in logging.

Reason: The race condition is throwing to_version as from_version during client upgrade task. Which is giving wrong status of client upgrade.

Impact: This will not have any impact on client upgrade or logging. Only the previos version of client (katello-agent) wont be displayed in Upgrade status Email. This is absolutely low severity removal of code.


jyejare:
closes #546 .

san7ket:
ACK

PR #558, opened 2017-04-04
Fix paramiko logging handler issue

jyejare:
We were getting paramiko errors while running fab tasks.

The error was something like this:
```
No handlers could be found for logger "paramiko.transport"

```

I disable the paramiko logging which wont impact on our upgrade.


san7ket:
ACK , when travis is done

PR #559, opened 2017-04-04
Add support for applying a hotfix on the Satellite6 Server.

kbidarkar:
The repos and the script should reside on the HTTP_SERVER_HOSTNAME
specified in the jenkins-config repo.

The repos and the script are unique depending upon the OS version.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1358,6 +1358,34 @@ def configure_realm(admin_password=None, keytab_url=None, realm=None,
     run('service foreman-proxy restart')
 
 
+def apply_hotfix():
+    """Apply a hotfix for the Satellite6 Server
+
+    Expects the following environment variables:
+
+    HOTFIX
+        Specifies the choice for the hotfix, the default or custom.
+    HTTP_SERVER_HOSTNAME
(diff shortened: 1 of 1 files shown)


kbidarkar:
So, while working on this PR, I came across a situation where in initially I thought we should allow the user to specify the repo and script, but soon realized that it would lead to unnecessary parameter clutter on the UI. Hence added a help-tip for it.

automation_tools/__init__.py
```diff
@@ -1358,6 +1358,34 @@ def configure_realm(admin_password=None, keytab_url=None, realm=None,
     run('service foreman-proxy restart')
 
 
+def apply_hotfix():
+    """Apply a hotfix for the Satellite6 Server
+
+    Expects the following environment variables:
+
+    HOTFIX
+        Specifies the choice for the hotfix, the default or custom.
```
sghai: So `Default` is when Hotfix is available through any repo and `custom` when its available through custom script ?

kbidarkar: Yes, that is right.

sthirugn:
ACK

sghai:
ACK

PR #560, opened 2017-04-04
Added support for running downstream-nightly.

kbidarkar:
Updated nightly to upstream-nightly and introduced new
satellite_version as downstream-nightly.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -348,7 +348,7 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         proxy = 'capsule'
     else:
         proxy = 'foreman-proxy'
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -348,7 +348,7 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         proxy = 'capsule'
     else:
         proxy = 'foreman-proxy'
-    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1', '6.2', '6.3'):
+    if os.environ.get('SATELLITE_VERSION') != 'upstream-nightly':
```
kbidarkar: Yes this would be required even for downstream-nightly and hence the change.

kbidarkar: I was wrong this is not required, will revert it back.

PR #565, opened 2017-04-05
downstream-nightly uses foreman-proxy prefix for puppet

kbidarkar:


sghai:
ACK

san7ket:
ACK

lpramuk:
if downstream-nightly uses foreman-proxy prefix for puppet then 6.3 will also use it 
then why it is in condition ?

kbidarkar:
probably yes, we may be 100% sure once official 6.3 builds arrive, but what I did in this PR was just revert it back to previous state.

Issue #570, opened 2017-04-07
Fix Collection of preupgrade and postupgrade datastore' without asking


jyejare:
Upgrade jobs are collecting preupgrade and postupgrade datastore for running existance tests without asking to do so.
Fix that.

jyejare:
Closing this issue as the ref. PR is merged.

PR #571, opened 2017-04-07
Fix unnecessary collection of pre and post upgrade datastore

jyejare:
This PR fixes the code that was collecting the pre and post upgrade datastore without asking to do so.

### DIFF
--- a/automation_tools/satellite6/upgrade/satellite.py
+++ b/automation_tools/satellite6/upgrade/satellite.py
@@ -66,7 +66,7 @@ def satellite6_setup(os_version):
     # Set satellite hostname in fabric environment
     env['satellite_host'] = sat_host
     logger.info('Satellite {} is ready for Upgrade!'.format(sat_host))
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if os.environ.get('RUN_EXISTANCE_TESTS', 'false').lower() == 'true':
         logger.info('Setting up preupgrade datastore for existance tests')
         set_datastore('preupgrade')
     return sat_host
@@ -168,7 +168,7 @@ def satellite6_upgrade():
     # Enable ostree feature only for rhel7 and sat6.2
     if to_version == '6.2' and major_ver == '7':
         enable_ostree(sat_version='6.2')
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if os.environ.get('RUN_EXISTANCE_TESTS', 'false').lower() == 'true':
         logger.info('Setting up postupgrade datastore for existance tests..')
         set_datastore('postupgrade')
 
@@ -256,6 +256,6 @@ def satellite6_zstream_upgrade():
     set_hammer_config()
     hammer('ping')
     run('katello-service status', warn_only=True)
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if os.environ.get('RUN_EXISTANCE_TESTS', 'false').lower() == 'true':
         logger.info('Setting up postupgrade datastore for existance tests')
         set_datastore('postupgrade')


jyejare:
closes #570

automation_tools/satellite6/upgrade/satellite.py
```diff
@@ -66,7 +66,7 @@ def satellite6_setup(os_version):
     # Set satellite hostname in fabric environment
     env['satellite_host'] = sat_host
     logger.info('Satellite {} is ready for Upgrade!'.format(sat_host))
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if os.environ.get('RUN_EXISTANCE_TESTS').lower() == 'true':
```
elyezer: Make sure to default to a string if the env var is not defined like:

```python
os.environ.get('RUN_EXISTANCE_TESTS', '')
```

Because if that env var is no defined it will return None (default value that get method returns) and calling `lower` on a None object raises and exception.

elyezer: Do this for all changes.

elyezer:
I've approved to not block on me, but please make sure to have the changes in place before merging this PR.

automation_tools/satellite6/upgrade/satellite.py
jyejare: This environment variable is from jenkins job parameters and hence it will be either false or true. This will be always defined.

elyezer: What if we migrate jenkins and for some reason that param is forgot to be added? That is the point I had in mind when suggesting this. Also if someone wants to test locally he/she may forget to set this up.

jyejare:
@elyezer , Addressed your comment. Please revisit.

automation_tools/satellite6/upgrade/satellite.py
```diff
@@ -66,7 +66,7 @@ def satellite6_setup(os_version):
     # Set satellite hostname in fabric environment
     env['satellite_host'] = sat_host
     logger.info('Satellite {} is ready for Upgrade!'.format(sat_host))
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if bool(os.environ.get('RUN_EXISTANCE_TESTS', False)):
```
elyezer: Doing this way RUN_EXISTENCE_TESTS will run even when == `false`.

Better to do this:

```python
if os.environ.get('RUN_EXISTANCE_TESTS', 'false').lower() == 'true':
   # ...
```

elyezer:
```python
>>> d = {'RUN_EXISTANCE_TESTS': 'false'}
>>> if bool(d.get('RUN_EXISTANCE_TESTS', False)):
...     print('running')
... 
running
```

While the below code works as expected:

```python
>>> d = {'RUN_EXISTANCE_TESTS': 'false'}
>>> if d.get('RUN_EXISTANCE_TESTS', 'false').lower() == 'true':
...     print('running')
... 
>>> d = {'RUN_EXISTANCE_TESTS': 'true'}
>>> if d.get('RUN_EXISTANCE_TESTS', 'false').lower() == 'true':
...     print('running')
... 
running
```

automation_tools/satellite6/upgrade/satellite.py
```diff
@@ -168,7 +168,7 @@ def satellite6_upgrade():
     # Enable ostree feature only for rhel7 and sat6.2
     if to_version == '6.2' and major_ver == '7':
         enable_ostree(sat_version='6.2')
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if os.environ.get('RUN_EXISTANCE_TESTS').lower() == 'true':
```
elyezer: This should be updated

automation_tools/satellite6/upgrade/satellite.py
```diff
@@ -256,6 +256,6 @@ def satellite6_zstream_upgrade():
     set_hammer_config()
     hammer('ping')
     run('katello-service status', warn_only=True)
-    if os.environ.get('RUN_EXISTANCE_TESTS'):
+    if os.environ.get('RUN_EXISTANCE_TESTS').lower() == 'true':
```
elyezer: This should be updated

PR #572, opened 2017-04-11
Yum lock issue fix by killing PID

jyejare:
In docker containers, due to multiple reasons the yum service was not been freed up to make it available for mext task. Due to which the client upgrade is failing while attempting to update katello-agent package.

So fixing this issue by killing this yum.pid before using the yum.

Also removing the earlier workaround of hard sleep times.

### DIFF
--- a/automation_tools/satellite6/upgrade/client.py
+++ b/automation_tools/satellite6/upgrade/client.py
@@ -135,16 +135,12 @@ def satellite6_client_setup():
             int(clients_count)/2,
             host=docker_vm
         )[docker_vm]
-        # Adding sleep to freed up yum lock taken by above client generate task
-        time.sleep(30)
         clients7 = execute(
             generate_satellite_docker_clients_on_rhevm,
             'rhel7',
             int(clients_count)/2,
             host=docker_vm
         )[docker_vm]
-        # Adding sleep to freed up yum lock taken by above client generate task
-        time.sleep(30)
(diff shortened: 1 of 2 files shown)


automation_tools/satellite6/upgrade/client.py
```diff
@@ -135,16 +135,12 @@ def satellite6_client_setup():
             int(clients_count)/2,
             host=docker_vm
         )[docker_vm]
-        # Adding sleep to freed up yum lock taken by above client generate task
-        time.sleep(30)
```
sghai: Good that we removed the sleeps here..

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -551,10 +551,15 @@ def refresh_subscriptions_on_docker_clients(container_ids):
         for container_id in container_ids:
             docker_execute_command(
                 container_id, 'subscription-manager refresh')
+            if docker_execute_command(
+                    container_id, '[ -f /var/run/yum.pid ]; echo $?') == '0':
+                docker_execute_command(container_id, 'rm -rf /var/run/yum.pid')
```
sghai: Lets see if it really fix the issue.

elyezer: I am wondering if we should also try to kill the process...

PR #573, opened 2017-04-11
Default GATEWAY and DHCP_RANGE to the foreman libvirt network

lpramuk:
- allow discovery to work outside of the lab (in standalone auto)


lpramuk:
Tested and discovery works

PR #575, opened 2017-04-12
Refactoring the product_upgrade job, to support client and capsule upgrade

san7ket:
When the refactoring of the product_upgrade was done, to create a separate setup task and product upgrade task , it broke capsule and clients upgarde.
This pull will it 
Will also raise a fiix in robotello-ci, to support this.

### DIFF
--- a/automation_tools/satellite6/upgrade/__init__.py
+++ b/automation_tools/satellite6/upgrade/__init__.py
@@ -25,6 +25,8 @@ from automation_tools.satellite6.upgrade.satellite import (
 from automation_tools.satellite6.upgrade.tasks import (
     get_sat_cap_version,
     post_upgrade_test_tasks,
+    create_setup_dict,
+    get_setup_data
 )
 from automation_tools.satellite6.upgrade.tools import logger
 from distutils.version import LooseVersion
@@ -86,11 +88,17 @@ def setup_products_for_upgrade(product, os_version):
     if product == 'client' or product == 'longrun':
         logger.info('Setting up Clients ....')
         clients6, clients7 = satellite6_client_setup()
+    setups_dict = {
+        'sat_host': sat_host,
+        'capsule_hosts': cap_hosts,
+        'clients6': clients6,
+        'clients7': clients7
+    }
+    create_setup_dict(setups_dict)
     return sat_host, cap_hosts, clients6, clients7
 
 
-def product_upgrade(
-        product, sat_host, cap_hosts, clients6=None, clients7=None):
+def product_upgrade(product):
     """Task which upgrades the product.
 
     Product is satellite or capsule or client or longrun.
@@ -102,14 +110,6 @@ def product_upgrade(
     z-stream released version
 
     :param string product: product name wanted to upgrade.
-    :param string sat_host: sat_host is satellite hostname.
-    :param string cap_hosts: cap_hosts is the capsule host's hostname.
-    :param string clients6: rhel 6 clients to be upgraded,
-        optional in case of docker clients.
-        If not given, the clients will be generated on docker.
-    :param string clients7: rhel 7 clients to be upgraded,
-        optional in case of docker clients.
-        If not given, the clients will be generated on docker.
 
     Environment Variables necessary to proceed Upgrade:
     -----------------------------------------------------
@@ -188,6 +188,9 @@ def product_upgrade(
         to_version = os.environ.get('TO_VERSION')
         logger.info('Performing UPGRADE FROM {0} TO {1}'.format(
             from_version, to_version))
+        # Get the setup dict returned by setup_products_for_upgrade
+        setup_dict = get_setup_data()
+        sat_host = setup_dict['sat_host']
         try:
             with LogAnalyzer(sat_host):
                 current = execute(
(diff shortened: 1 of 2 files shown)


san7ket:
also see: https://github.com/SatelliteQE/robottelo-ci/pull/588

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -86,11 +88,17 @@ def setup_products_for_upgrade(product, os_version):
     if product == 'client' or product == 'longrun':
         logger.info('Setting up Clients ....')
         clients6, clients7 = satellite6_client_setup()
+        global_dict = {
```
jyejare: Take this out of if loop of client, else this will never execute out in case of satellite and capsule upgrade.

automation_tools/satellite6/upgrade/__init__.py
```diff
@@ -86,11 +88,17 @@ def setup_products_for_upgrade(product, os_version):
...
+            'sat_host': sat_host,
+            'capsule_hosts': cap_hosts,
+            'clients6': clients6,
+            'clients7': clients7
+        }
+        create_setup_dict(global_dict)
```
jyejare: Same as above.

automation_tools/satellite6/upgrade/__init__.py
jyejare: Rename 'global_dict' to 'setups_dict' else a third person wont understand, what are the further uses of it.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -776,3 +778,20 @@ def csv_reader(component, subcommand):
         entity_list.append(row)
     comp_dict[component] = entity_list
     return comp_dict
+
+
+def create_setup_dict(dict):
```
jyejare: Rename 'dict' parameter to something else but not dict. That keyword is already reserved by python.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -44,6 +45,7 @@
 
 
 logger = logger()
+data = {}
```
jyejare: This is not required as we are directly passing a dictinary to json. Need not to initialize it.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -776,3 +778,20 @@ def csv_reader(component, subcommand):
         entity_list.append(row)
     comp_dict[component] = entity_list
     return comp_dict
+
+
+def create_setup_dict(dict):
+    data.update(dict)
```
jyejare: Write docstring for a function.

jyejare: Remove this as we are removing data variable. Not required as we are directly passing dictionary to json dump.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -776,3 +778,20 @@ def csv_reader(component, subcommand):
...
+    print data
```
jyejare: Remove this print statement.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -776,3 +778,20 @@ def csv_reader(component, subcommand):
...
+    with open('{}.pkcl'.format('product_setup'), 'wb') as pref:
```
jyejare: Its not required to keep its extension as pkcl. You can have a plain file without extension. Its non-blocer comment.

jyejare: You can directly keep the name as 'product_setup.pkcl', need not format the string.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -776,3 +778,20 @@ def csv_reader(component, subcommand):
...
+        pickle.dump(data, pref)
+
+
+def get_setup_data():
+    with open('{}.pkcl'.format('product_setup')) as pref:
```
jyejare: Same as above.

automation_tools/satellite6/upgrade/tasks.py
```diff
@@ -776,3 +778,20 @@ def csv_reader(component, subcommand):
...
+        data = pickle.load(pref)
+    clients6 = data['clients6']
+    clients7 = data['clients7']
+    sat_host = data['sat_host']
+    cap_hosts = data['capsule_hosts']
+    return sat_host, cap_hosts, clients6, clients7
```
jyejare: Directly return whole dictinstead of assigning them to variable as -
```
return data
```
But handle that while calling this function.

san7ket: Won't this be a more cleaner way, as this function just open's file to read a dict, adding the extraction of values here , itself will be good. Could I just add a docstring for function to explain what's it doing.

automation_tools/satellite6/upgrade/tasks.py
san7ket: This is initialize to create a empty dict, and then update it always. But yes, i will declare this locally into the function.

automation_tools/satellite6/upgrade/tasks.py
jyejare: If you are returning 4 values then at receiving end you always need 4 variables. And that is not in case of dictionary return.
Also the purpose of dictionary is being comprised here.

jyejare: You can return dict and ask for any single value in dict by its key. Also this is good coding standard.

jyejare: You can look at the code of ```create_datastore and get_datastore```.

PR #578, opened 2017-04-13
yum PID file removal leading to issues w/ client upgrade

sghai:
not sure if it really fixes the issue w/ current job but in any case removing pid file leads some other issues..


san7ket:
Changes looks good, this won't stop the whole job , if one of the clients fail to register , due to yum fix

PR #579, opened 2017-04-13
Quiet yum clean all ouput to fix yum lock issue

jyejare:
This quiets the output of yum clean all during client upgrade process to avoid the yum lock issue.


sghai:
ACK

PR #583, opened 2017-04-24
Enable satellite6 package to be accessible with pip

jyejare:
Currently only automation_tools package is listed to be accessible via pip.

This PR enables satellite6 package under automation_tools package to accessible via pip install.

This resolves the dependency issue with https://github.com/SatelliteQE/satellite6-upgrade.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
     author=u'Elyézer Rezende',
     author_email='erezende@redhat.com',
     url='https://github.com/SatelliteQE/automation-tools',
-    packages=['automation_tools'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
     author=u'Elyézer Rezende',
     author_email='erezende@redhat.com',
     url='https://github.com/SatelliteQE/automation-tools',
-    packages=['automation_tools'],
+    packages=['automation_tools', 'automation_tools/satellite6'],
```
sghai: 👍

sghai:
ACK

PR #586, opened 2017-04-28
Added the setup part required for code coverage

kbidarkar:


automation_tools/__init__.py
```diff
@@ -628,6 +628,37 @@ def setup_abrt():
...
+    sitecustomize_file = '/usr/lib/python2.7/site-packages/sitecustomize.py'
+    sitecustomize_url = os.environ.get('SITECUSTOMIZE_URL')
+    # Install EPEL packages for the installation
+    run('yum -y localinstall https://dl.fedoraproject.org'
+        '/pub/epel/epel-release-latest-{0}.noarch.rpm'
+        .format(os_version))
+    run('yum -y install python-pip')
+    run('pip install -U coverage')
+    # Delete EPEL repo files
+    delete_custom_repos('epel*')
```
omaciel: 👍

automation_tools/__init__.py
```diff
@@ -628,6 +628,37 @@ def setup_abrt():
...
+
+    Expects the following environment variables:
+
+        SITECUSTOMIZE_URL
+                The sitecustomize file to launch the coverage.
+        COVERAGERC_URL
+                The coverage configuration file.
+
+    """
+    coverage_file = '/etc/coverage/.coveragerc'
```
omaciel: Since this is a constant, perhaps rename it to `COVERAGE_FILE` and put it together with other constants?

automation_tools/__init__.py
```diff
@@ -628,6 +628,37 @@ def setup_abrt():
...
+    coverage_url = os.environ.get('COVERAGERC_URL')
+    os_version = distro_info()[1]
+    sitecustomize_file = '/usr/lib/python2.7/site-packages/sitecustomize.py'
```
omaciel: See my comment about constants.

automation_tools/__init__.py
elyezer: I think it is ok on leaving here, having this value on a variable improve the code readability and having them here will avoid need to check on other part of the code for the value. 

I would suggest moving to a constants place if this value was used in more than one function.

automation_tools/__init__.py
elyezer: Same as https://github.com/SatelliteQE/automation-tools/pull/586/files#r114414073

automation_tools/__init__.py
```diff
@@ -1910,6 +1941,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
             sat_version=satellite_version,
             host=host
         )
+    execute(setup_code_coverage, host=host)
```
elyezer: Will the services be restarted after this? They need to be restarted otherwise coverage data collection won't work.

kbidarkar: There is no need to restart the services at this stage as with all Tiers , it's all instance based, so anyways it would get started with a new instance for each Tier.

PR #590, opened 2017-05-08
Remove epel package if installed and then install the epel package.

kbidarkar:
It is better to uninstall epel package rather than deleting the epel.repo file, so that proper usage of epel repo is possible once again in the provisioning job stage.


sthirugn:
Thank you @kbidarkar ACK

PR #597, opened 2017-05-18
Move existing ssh keys to a bkp dir, to avoid any mismatch of key pair

kbidarkar:
This is required to avoid the below issues, seen from the log,
https://jenkins.redhat.com/view/Satellite%206/job/satellite6-installer/2670/console

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -859,7 +859,11 @@ def setup_libvirt_key():
     if key_url is None or libvirt_host is None:
         print('You must specify the Libvirt key URL and the Libvirt hostname')
         sys.exit(1)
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -859,7 +859,11 @@ def setup_libvirt_key():
     if key_url is None or libvirt_host is None:
         print('You must specify the Libvirt key URL and the Libvirt hostname')
         sys.exit(1)
-    # Deploy priv ssh key under root and foreman account
+    # Move existing root user ssh keys, to avoid the ssh key pair mismatch.
```
sghai: 👍

PR #598, opened 2017-05-19
Setup code coverage to run only for provisioning jobs.

kbidarkar:
This is done, so that code_coverage is not configured while
configuring a setup via the satellite6 installer job.


sghai:
ACK

PR #601, opened 2017-05-23
run code_coverage only for RHEL7 OS

kbidarkar:


automation_tools/__init__.py
```diff
@@ -1996,7 +1999,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
             host=host
         )
     # Setup code_coverage only for the provisoning jobs.
-    if 'base' in target_image:
+    if 'TARGET_IMAGE' in os.environ and 'base' in target_image:
```
omaciel: This part is not intuitive to me. Why is this change needed?

kbidarkar: This is required, as the same code is used for both `Satellite6 Installer` and `provisioning jobs`, 
for `Satellite6 Installer` , `TARGET_IMAGE` is never exported and hence it may fail.

PR #606, opened 2017-05-25
Increase the timeout to 120 seconds as at times it's slow.

kbidarkar:


kbidarkar:
Guys, will merge this, I need this as provisioning on pap**** is failing as it takes atleast 100 secs for the vm to be up.

rplevka:
could you possibly adapt the approach we use in `robottelo.vm`?:
https://github.com/SatelliteQE/robottelo/blob/master/robottelo/vm.py#L178

PR #608, opened 2017-05-29
Add support to configure uploading of coverage results to sonarqube

kbidarkar:


automation_tools/__init__.py
```diff
@@ -689,6 +689,35 @@ def setup_code_coverage():
...
 
 
+def configure_sonarqube():
+    """Task to configure SonarQube."""
+    http_server = os.environ.get('HTTP_SERVER_HOSTNAME')
+    sonar_server = os.environ.get('SONAR_SERVER_URL')
+    satellite_version = os.environ.get('SATELLITE_VERSION')
+    build_label = os.environ.get('BUILD_LABEL')
+    # Sonar Scanner requires Java1.8 to be able to upload content to SonarQube
+    run('yum -y install java-1.8.0-openjdk')
```
kbidarkar: We will be installing java-1.8.0-openjdk for the client side of the SonarQube which is Sonar Scanner. 
Also the task `configure_sonarqube` will only be run on the `coverage` instance and hence upgrading java here outside of satellite packages should be ok, as `coverage` instance runs no testcases.

PR #609, opened 2017-05-31
Skip pytest test if bugzilla/redmine bug is open

jyejare:
The decorator to skip the test if bugzilla / redmine bug is open.

This decorator will be mostly used by SAtellite6-upgrade tests as of now.

### DIFF
--- a/automation_tools/bz.py
+++ b/automation_tools/bz.py
@@ -5,6 +5,7 @@ copied from robottelo's robotello/decorators/__init__.py
 """
 import bugzilla
 import logging
+import pytest
 import requests
 
 from six.moves.xmlrpc_client import Fault
@@ -35,6 +36,10 @@ _redmine = {
 
 class BugFetchError(Exception):
     """Indicates an error occurred while fetching information about a bug."""
+
+
+class BugTypeError(Exception):
+    """Indicates that an incorrect bug type was specified."""
 
 
 def _get_bugzilla_bug(bug_id):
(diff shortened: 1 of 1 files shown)


elyezer:
Duplicating code is always a sign that something must be built to avoid that. Creating a package with that is a good idea.

jyejare:
@elyezer , @renzon @rochacbruno , So Why dont we make automation-tools repo to be used by all projects like robotello, upgrade and anything else.
We will dump everything here to be used by all. Or We should create a common repository that will be used by all. What do you think guys?

elyezer:
> So Why dont we make automation-tools repo to be used by all projects like robotello, upgrade and anything else.

Automation tools requires fabric to be installed and that is not a dependency that would make sense for other projects and requiring it to be installed seems not to be reasonable. Also the automation-tools structure is not the best in terms of a Python package.

> We will dump everything here to be used by all.

Some things were already dumped here and it started to become a Frankenstein. Automation-tools purpose is to offer frabic tasks to setup the infrastructure we need for our tests, just that. It will eventually be replaced by some ansible playbooks.

Having a separated project that offers helpers for inspecting Bugzilla and redmine is something that can be used not only by us but by other teams and I don't think they will like to inherit satellite related configuration tasks just to have that. I think we can take this opportunity and create a tool that will make others aware of our work.

rochacbruno:
Waiting the revire, merge and PyPI release of https://github.com/SatelliteQE/robozilla/pull/28 

to unblock

@jyejare once the PR https://github.com/SatelliteQE/robozilla/pull/28 is merged and `robozilla 0.1.8` released to PyPI you will be able to do.

```python
from robozilla.decorators import bz_bug_is_open
```

optionally it accepts 

`sat_version_picker=function_returning_string` 

and `config_picker=function_returning_a_dict_with_bz_credentials`

Take a look at robozilla.decorators docstrings once merged.

jyejare:
The implementation is already done in robozilla as per above comment. And hence closing this PR.

PR #611, opened 2017-06-05
Fix typo: s/upstream-nightly/nightly/

lpramuk:
Fixes upstream provisioning failure:
```
[server] out: ERROR: No installation scenario was selected, the installer cannot continue.
[server] out:        Even --help content is dependent on selected scenario.
[server] out:        Select scenario with --scenario SCENARIO or list available scenarios with --list-scenarios.
[server] out: 

Fatal error: run() received nonzero return code 26 while executing!
```


lpramuk:
oops typo was in my mind only

PR #615, opened 2017-06-16
Hotfix: lsmod fails in a VM

lpramuk:
hotfix for:
```
Fatal error: run() received nonzero return code 1 while executing!

Requested: lsmod | grep kvm_
Executed: /bin/bash -l -c "lsmod | grep kvm_"

Aborting.

[server] run: lsmod | grep kvm_
```


lpramuk:
This fail occurred in upstream nightly which is weird - it tries to setup virtualization in a VM

PR #617, opened 2017-06-19
Open dhcpd port (67/udp) only for internal capsule within lab

### LINKED ISSUE #616: Open dhcpd port 67/udp on Satellite in automation lab
Provisioning using internal capsule in automation never worked.
Guess why? Dhcpd service wasn't ever exposed.

But in standalone installer we don't want to expose this port as we would mess up beaker lab.

lpramuk:
- moved setup_satellite_firewall() to run earlier (katellovirt module no longer present)
- run aditional non-flushing setup_firewall() only for internal capsule in the lab to add ```67/udp```
- simplified logic wrapped around setup_default_capsule() (setup_default_libvirt() moved up)

(Fixes #616)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -324,8 +324,6 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         sys.exit(1)
 
     if interface is None:
-        # Setup libvirt and fetch virtual bridge information
-        interface = setup_default_libvirt()
         if len(interface) == 0:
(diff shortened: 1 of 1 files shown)


rplevka:
This is needed to unblock the VM provisioning with bridge ifaces on our external libvirt CR.
The current situation blocks DHCP traffic while no interface is specified to installer (the scenario we have in a lab), so VMs are unable to pick up an IP,  thus provisioning gails.

PR #619, opened 2017-06-21
Allow the partition_disk job to successfully run on RHEL 6

### LINKED ISSUE #618: partition_disk fails on rhel 6
When running this on a rhel 6 host, we get the following output 

```
+ fab -H root@host partition_disk
[root@host] Executing task 'partition_disk'
[root@host] run: stat --format=%m /home
[root@host] out: ?
[root@host] out: 


Done.
Disconnecting from root@host... done.
```

It turns out that the %m option is not available in stat for at least RHEL 6.8+. One alternative, would be a command like this, that has the same output, but works on both RHEL 6 and RHEL 7.

```
-bash-4.1# df -P /home | awk 'END{print $NF}'
/home
```

JacobCallahan:
On RHEL 6.8+, the %m format is not available for the stat command.
Switching to this version, which produces the same output, and works
on both RHEL 6 and RHEL 7.

Fixes #618

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2115,7 +2115,7 @@ def partition_disk():
     synchronization of larger repositories.
 
     """
-    if run('stat --format=%m /home') == '/home':
+    if run('df -P /home | awk \'END{print $NF}\'') == '/home':
         run('umount /home')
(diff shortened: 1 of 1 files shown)


JacobCallahan:
My tests of this change can be found in the linked issue.

PR #620, opened 2017-06-27
Move fix_qdrouterd_listen_to_ipv6() task after all installer runs

lpramuk:
... so that changes are not overridden by a successive installer run

The task changes were not effective (still listening on ipv4 only) as we are running installer in postinstall time once more (enable_ostree ) since Oct 2016

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2046,7 +2046,6 @@ def product_install(distribution, create_vm=False, certificate_url=None,
             certificate_url,
             host=host
         )
-    execute(fix_qdrouterd_listen_to_ipv6, host=host)
     execute(setup_alternate_capsule_ports, host=host)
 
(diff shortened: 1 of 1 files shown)


lpramuk:
You have ```/etc/qpid-dispatch/qdrouterd.conf``` containing
```
listener {
    addr: ::
    port: 5647
    sasl-mechanisms: ANONYMOUS
    ssl-profile: server
}
```
after any installer run you get 
```
listener {
    addr: 0.0.0.0
    port: 5647
    sasl-mechanisms: ANONYMOUS
    ssl-profile: server
}
```
That' why this PR moves ```fix_qdrouterd_listen_to_ipv6()``` task to be run later

sghai:
ACK

rplevka:
ACK
thanks for addressing this

PR #629, opened 2017-10-03
Fix Capsule cert generate fqdn option in 6.3

jyejare:
This fixes the capsule cert generate option in 6.3


sghai:
ACK.. looks good.

PR #630, opened 2017-10-03
Capsule Cert Generate Absolute path w/o

jyejare:
w/o for bug https://bugzilla.redhat.com/show_bug.cgi?id=1466688

### DIFF
--- a/automation_tools/satellite6/capsule.py
+++ b/automation_tools/satellite6/capsule.py
@@ -15,6 +15,7 @@ from hammer import (
     hammer_repository_synchronize,
     set_hammer_config
 )
+from robozilla.decorators import bz_bug_is_open
 
 
 class ImproperlyConfigured(Exception):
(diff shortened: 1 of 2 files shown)


automation_tools/satellite6/capsule.py
```diff
@@ -206,7 +207,11 @@ def generate_capsule_certs(capsule_hostname, force=False):
         fqdn_opt = '--foreman-proxy-fqdn'
     else:
         fqdn_opt = '--capsule-fqdn'
-    cert_path = '{0}-certs.tar'.format(capsule_hostname)
+    if bz_bug_is_open(1466688):
+        # Absolute path bug
+        cert_path = '~/{0}-certs.tar'.format(capsule_hostname)
```
sghai: this may not work because of another bug: 1470653

jyejare: We are not double quoting the cert_path string so it should work

sghai: Looks like automation won't include quotes.. in that case it will work.

sghai:
ACK

PR #631, opened 2017-10-10
Setup default subnet in postinstall time

lpramuk:
Satellite installer currently has options only to setup default domain (we do it) but has none to setup default subnet to match Internal Capsule network setup. 

So, we need to set up some "default subnet" in postinstall time unless there are installer options for it.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -433,6 +433,39 @@ def setup_default_libvirt(bridge=None):
     return interface
 
 
+def setup_default_subnet():
+    """Postinstall task to setup default subnet within Satellite
+
+    Expects the following environment variables:
+
+    SUBNET
+        The network address that Internal Capsule resides in
+    NETMASK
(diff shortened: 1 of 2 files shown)


rplevka:
I've used a different approach in my commit -
I simply create a subnet on-demand directly in the test by sourcing the vlan settings from `robottelo.settings.vlan_networking`.
Satellite allows us to create subnets with identical settings (networks, masks, gateways, etc) as long as they have unique name.

Wouldn't be that more robust? my way, the tests would be independent on the setup and also they would not suffer if some other careless tests delete the subnet for some selfish reason.
We can also combine these two approaches and use the runtime subnet creation only if the desired subnet does not exist yet.

https://github.com/rplevka/robottelo/commit/71a4d9f5431c1c534c7f4a7e11de3c76ae1f8d2d

lpramuk:
we can have both, default subnet is useful even for manual testing

PR #640, opened 2017-11-09
Added vnc-server service and it's port-range

kbidarkar:


automation_tools/__init__.py
```diff
@@ -428,6 +428,12 @@ def setup_default_libvirt(bridge=None):
         run('virsh net-start foreman')
     run('virsh net-autostart foreman')
 
+    os_version = distro_info()[1]
+    if os_version >= 7:
```
omaciel: Unless I'm missing something (sorry, I haven't looked at this repo in a while), this would only configure the firewall if the OS is RHEL 7 or higher... why not also handle it for RHEL 6?

lpramuk: there is "else:" clause yet, but you probably already noticed :)

kbidarkar: Earlier we had, only for RHEL7 later added for RHEL6 too and hence the else now.

PR #641, opened 2017-11-20
Fix setup_foreman_discovery and setup_default_subnet for upstream-nig…

kbidarkar:
…htly


automation_tools/__init__.py
```diff
@@ -476,7 +476,7 @@ def setup_default_subnet(sat_version):
         '--gateway {gateway} --dns-primary {gateway} '
         '--ipam DHCP --from {from} --to {to} '
         '--dhcp-id 1 --dns-id 1 --tftp-id 1 ' +
-        ('--discovery-id 1' if sat_version > '6.2' else '')
```
kbidarkar: This is updated as sat_version can also be a string like `upstream-nightly` or `downstream-nightly`

PR #643, opened 2018-01-09
Added task for creating discovery image

ntkathole:
This task will create foreman-discovery-image depending on vlan under ` /var/lib/libvirt/images/`  depending on environment variables` BASE_URL`, `DISCOVERY_ISO`, `GATEWAY` and `IPADDR`

### DIFF
--- a/automation_tools/utils.py
+++ b/automation_tools/utils.py
@@ -1,9 +1,12 @@
 """Utilities tasks and functions"""
 from __future__ import print_function
 
+import os
 import re
 import sys
+import urllib2
 
+from BeautifulSoup import BeautifulSoup
 from fabric.api import env, run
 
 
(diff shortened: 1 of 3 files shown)


automation_tools/utils.py
```diff
@@ -79,3 +82,36 @@ def update_packages(*args, **kwargs):
...
+        subprocess.call(['pip', 'install', 'BeautifulSoup'])
+        from BeautifulSoup import BeautifulSoup
+
+    soup = BeautifulSoup(urllib2.urlopen(url).read())
+    for link in soup.findAll('a'):
+        if 'foreman-discovery-image' in link.string:
+            discovery_image = link.string
+    try:
+        run("wget -O /tmp/" + discovery_image + " " + url + discovery_image)
+        run('rpm -Uvh /tmp/' + discovery_image)
```
sghai: just thinking instead of installing rpm, lets try to extract the files through rpm2cpio, e.g. `cd tmpdir;  rpm2cpio discovery.rpm | cpio -idmv`

sghai:
ty @ntkathole for working on this.

automation_tools/utils.py
```diff
@@ -79,3 +82,37 @@ def update_packages(*args, **kwargs):
...
+
+def get_discovery_image():
+    """ Task for getting unattended foreman-discovery ISO image
+    :return: foreman-discovery-image iso under /var/lib/libvirt/images/
+    """
+    url = os.environ.get('BASE_URL') + '/Packages/'
+    try:
+        from BeautifulSoup import BeautifulSoup
+    except Exception:
+        subprocess.call(['pip', 'install', 'BeautifulSoup'])
```
jyejare: Instead, you add this is requirements, which will install BeautifulSoap for sure. And then Remove try block

ntkathole: @jyejare yes I thought of that. But other side is why to add extra requirement to whole project (causing to install it everytime thought it is not required) for just one task. What do you think?

jyejare: True ! +1

kbidarkar: AFAIK, imports into a local scope, such as inside a function definition should be avoided, people prefer all the imports to happen at the top so that anyone can quickly know what are the dependencies.

kbidarkar: @ntkathole , would request an update to the PR and move the import statement for `BeautifulSoap` to the top of the
module.

kbidarkar: Also, the `BeautifulSoup` as a dependency should go inside `install_requires` under setup.py

https://github.com/SatelliteQE/automation-tools/blob/master/setup.py#L31

kbidarkar: As to why it should be avoided is a good question, but with my experience, I say it is not a good practice to handle installation of the modules inside the code. 

Maybe this link helps, https://stackoverflow.com/questions/6120902/how-do-i-automatically-install-missing-python-modules?noredirect=1

kbidarkar:
@ntkathole, beaker-builder version 1.9 is now installed on Jenkins and available.

ntkathole:
@kbidarkar Thanks for review! comments addressed.

automation_tools/utils.py
```diff
@@ -1,8 +1,11 @@
 """Utilities tasks and functions"""
 from __future__ import print_function
+from BeautifulSoup import BeautifulSoup
```
kbidarkar: Last comment, 
1) This line should go before Line10.
2) `from __future__ import print_function`  is an exception as, the whole point of `from __future__ import print_function` is to bring the print function from Python 3 into Python 2.6+.

setup.py
```diff
@@ -21,6 +21,7 @@
     package_dir={'automation_tools': 'automation_tools'},
     include_package_data=True,
     install_requires=[
+        'BeautifulSoup',
```
kbidarkar: 👍

ntkathole:
@kbidarkar thanks! updated.

kbidarkar:
ACK

PR #648, opened 2018-02-20
6.3 GA Update

JacobCallahan:
This set of changes should allow us to install against 6.3 GA.

I may have missed something, but hopefully this will at least get us going!


cswiii:
ACK

PR #651, opened 2018-02-21
Fix too more changes for 6.3 GA

lpramuk:
What we do for 6.1 and 6.2 when setting up foreman discovery
we don't want to do for 6.3

Fix of changes brought by PR#649

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -866,7 +866,7 @@ def setup_foreman_discovery(sat_version):
     )
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
-    if sat_version in ('6.1', '6.2', '6.3'):
+    if sat_version in ('6.1', '6.2'):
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -866,7 +866,7 @@ def setup_foreman_discovery(sat_version):
     )
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
-    if sat_version in ('6.1', '6.2', '6.3'):
+    if sat_version in ('6.1', '6.2'):
```
rplevka: you are correct. the 6.3 case is handled at
https://github.com/SatelliteQE/automation-tools/pull/651/files#diff-a692ae115cc45cc7627d4545e9b4c294R906
and it differs from the older versions. Great catch, btw.

JacobCallahan:
ACK. Wasn't sure on this one, but we needed to get the 6.3 GA stuff running. Good catch, @lpramuk

PR #653, opened 2018-02-22
Achieve python3 compatibility while maintaining python2.7 compatibility

lpramuk:
Tested on provisioning upstream job with CPython3.5 Builder


lpramuk:
I better test also  CPython2.7 Builder ....

Issue #654, opened 2018-02-23
[Prio 0] Account for different python string types during setup_code_coverage


JacobCallahan:
During our latest RHEL 7 ISO installation attempt, the installer ran into this issue.

```
run: mkdir -p /etc/coverage
Traceback (most recent call last):
...
  File "/home/jenkins/workspace/provisioning-6.2-rhel7/automation_tools/__init__.py", line 737, in setup_code_coverage
    coveragerc_file.write('[run]\n')
TypeError: unicode argument expected, got 'str'
```

I did some quick searching and found a possible solution

```python
try:
    # Python 2
    from cStringIO import StringIO
except ImportError:
    # Python 3
    from io import StringIO
```
Then remove ```io.``` from the rest of the code base.

JacobCallahan:
Looking at our code, it looks like that solution above probably won't work, since we try to import from essentially the same module.

kbidarkar:
Fixed in #655

PR #655, opened 2018-02-23
Fix TypeError: unicode argument expected, got 'str'

lpramuk:
Root cause:
```
$ python2
Python 2.7.14 (default, Dec 11 2017, 16:08:01) 
[GCC 7.2.1 20170915 (Red Hat 7.2.1-2)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from io import StringIO
>>> StringIO().write('str')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unicode argument expected, got 'str'
>>> StringIO().write(u'str')
3L
```

I switched to start using StringIO from io and forgot to add u'str'


JacobCallahan:
@lpramuk if you can't get any more reviewers in the next 30mins, I'll check back in and merge this.

Fixes #654

PR #657, opened 2018-03-05
Reverse sat_version logic when appending '--scenario' option

lpramuk:
Reverse sat_version logic when appending ```--scenario``` option
- simply in 6.0, 6.1 there was no option --scenario, starting 6.2 there always is
- helps with future versions 6.4+ or upstream-koji or whatever we would add


kbidarkar:
ACK

PR #665, opened 2018-03-09
Add extra rpms and installer options for upstream to match downstream

lpramuk:
- adds option to use koji repos for upstream (SATELLITE_VERSION=koji)
- enables REX for upstream (disabled by default)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1741,9 +1741,10 @@ def apply_hotfix():
         run('/root/hotfix.sh')
 
 
-def upstream_install(admin_password=None, run_katello_installer=True,
-                     puppet4=True):
+def upstream_install(admin_password=None, run_katello_installer=True):
     """Task to install Foreman nightly using forklift scripts"""
+    use_koji = 'koji' in os.environ.get('DISTRIBUTION').lower()
+    puppet4 = bool(os.environ.get('PUPPET4_REPO'))  # otherwise puppet5
     if admin_password is None:
         admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
@@ -1757,13 +1758,14 @@ def upstream_install(admin_password=None, run_katello_installer=True,
     run('git clone -q https://github.com/theforeman/forklift.git')
 
     with cd('forklift'):
-        # For now, puppet-four is configured by default. Toggle option
-        # will be added later.
         run('ansible-playbook -c local -i,$(hostname) '
-            '-e katello_version=nightly {0} '
+            '-e katello_version=nightly {0} {1} {2} '
             '-e foreman_installer_skip_installer=True '
             'playbooks/katello.yml'.format(
-                '-e puppet_repositories_version=4' if puppet4 else ''))
+                '-e puppet_repositories_version=4' if puppet4 else '',
+                '-e foreman_repositories_use_koji=True' if use_koji else '',
+                '-e katello_repositories_use_koji=True' if use_koji else '',
+            ))
 
     # Install support for various compute resources in upstream
     compute_resources = [
@@ -1775,7 +1777,18 @@ def upstream_install(admin_password=None, run_katello_installer=True,
         'rackspace',
         'vmware',
     ]
-    run('yum install -y foreman-{{{0}}}'.format(','.join(compute_resources)))
+    # Install hammer plugins to match downstream set of plugins
+    hammer_plugins = [
+        'csv',
+        'foreman_admin',
+        'foreman_openscap',
+        'foreman_remote_execution',
+        'foreman_virt_who_configure',
+    ]
+    run('yum install -y foreman-{{{0}}} tfm-rubygem-hammer_cli_{{{1}}}'.format(
+        ','.join(compute_resources),
+        ','.join(hammer_plugins),
+    ))
 
     installer_options = {
         'foreman-admin-password': admin_password,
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1741,9 +1741,11 @@ def apply_hotfix():
         run('/root/hotfix.sh')
 
 
-def upstream_install(admin_password=None, run_katello_installer=True,
-                     puppet4=True):
+def upstream_install(admin_password=None, run_katello_installer=True):
```
lpramuk: we have to have same signature for all ```*_install()``` puppet4 param couldn't be provided anyway, so removing

automation_tools/__init__.py
```diff
@@ -1741,9 +1741,11 @@ def apply_hotfix():
...
     """Task to install Foreman nightly using forklift scripts"""
+    sat_version = os.environ.get('SATELLITE_VERSION')
+    use_koji = 'koji' in sat_version
```
lpramuk: SATELLITE_VERSION=koji -> install UPSTREAM out of koji (more recent but can be broken or fixed already, depends on nightly health)

automation_tools/__init__.py
```diff
@@ -2867,23 +2875,22 @@ def katello_installer(debug=False, distribution=None, verbose=True,
...
-    if sat_version in ('6.0', '6.1'):
+    extra_options = []
+
+    if distribution == 'satellite6-upstream':
+        proxy = 'foreman-proxy'
+        installer = 'foreman'
+        scenario = 'katello'
+        extra_options.append('--enable-foreman-plugin-remote-execution')
+        extra_options.append('--enable-foreman-proxy-plugin-remote-execution-ssh')  # noqa
+    elif sat_version in ('6.0', '6.1'):
```
lpramuk: replaced independent
```
if 
if
```
with more proper chained ifs
```
if
elseif
```

automation_tools/__init__.py
```diff
@@ -2867,23 +2875,22 @@ def katello_installer(debug=False, distribution=None, verbose=True,
...
         proxy = 'capsule'
         installer = 'katello'
-    if distribution == 'sam-upstream':
```
lpramuk: sam no longer exists, so removing

automation_tools/__init__.py
```diff
@@ -2867,23 +2875,22 @@ def katello_installer(debug=False, distribution=None, verbose=True,
...
     # different values
-    if sat_version in ('6.0', '6.1'):
+    extra_options = []
+
+    if distribution == 'satellite6-upstream':
+        proxy = 'foreman-proxy'
+        installer = 'foreman'
+        scenario = 'katello'
+        extra_options.append('--enable-foreman-plugin-remote-execution')
+        extra_options.append('--enable-foreman-proxy-plugin-remote-execution-ssh')  # noqa
```
lpramuk: Add REX for upstream, by default is disabled

automation_tools/__init__.py
kbidarkar: The values of SATELLITE_VERSION are fixed, it can only be the below:
```
6.1
6.2
6.3
6.4
upstream-nightly
downstream-nightly
```
What we do currently, for various INTERNAL builds is specify it under SATELLITE_DISTRIBUTION

```
INTERNAL
INTERNAL AK
INTERNAL REPOFILE
```
We would need to take similar approach here.

automation_tools/__init__.py
```diff
@@ -1775,7 +1778,18 @@ def upstream_install(admin_password=None, run_katello_installer=True,
...
+        'csv',
+        'foreman_admin',
+        'foreman_openscap',
+        'foreman_remote_execution',
+        'foreman_virt_who_configure',
+    ]
+    run('yum install -y foreman-{{{0}}} tfm-rubygem-hammer_cli_{{{1}}}'.format(
+        ','.join(compute_resources),
+        ','.join(hammer_plugins),
+    ))
```
kbidarkar: This would is a good stuff to have, :+1:

lpramuk:
@kbidarkar 
 I think of it like this : UPSTREAM = { nightly, koji } that is ```SATELLITE_VERSION=koji```

With your propose to have ```DISTRIBUTION=koji``` what would be in SATELLITE_VERSION ? 

Every if concerning upstream would then turn into
```
if DISTRIBUTION==upstream or DISTRIBUTION=koji: 
```
instead of just simple:
```
if DISTRIBUTION==upstream:
```
I dont want to go into this changes

rplevka:
@kbidarkar - Frankly, I agree with @lpramuk - I don't see a reason of flattening these options. koji is pretty much the same as nightly, it's just not yet tested. Also, it's not really a distribution. I would treat it as a subcat of nightly.

automation_tools/__init__.py
lpramuk: I plan to add 'koji' in robottelo-ci later
this PR is not breaking anything without robottelo-ci changes

lpramuk: I can get `6.3` using INTERNAL or INTERNAL AK and should have the same result only means to obtain the same version are different. 

While with koji and nightly also the result is different so not only the means to get that.
That why we have to treat them as a separate versions (so not distributions)

kbidarkar:
```
SATELLITE_VERSION=upstream-nightly and SATELLITE_DISTRIBUTION=KOJI
SATELLITE_VERSION=upstream-nightly and SATELLITE_DISTRIBUTION=UPSTREAM
```
, that is what I am suggesting.

lpramuk:
@kbidarkar you've convinced me :-)
Since when we generate jobs we iterate over versions and I don't plan to have a separate job for koji.
I'm gonna treat koji as another DISTRIBUTION

lpramuk:
Comments resolved.
```DISTRIBUTION = koji```

**Comment by lpramuk** [comment]
Removed sam.

File: automation_tools/__init__.py [hunk 7]
```diff
@@ -2063,14 +2054,14 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         test_in_stage = (test_in_stage.lower() == 'true')
 
     install_tasks = {
-        'sam-upstream': sam_upstream_install,
```
kbidarkar: sam is being dropped as we never had a sam install for a while now.

kbidarkar:
ACK

PR #666, opened 2018-03-15
Forklift renamed koji options

lpramuk:
Original options:
- ```katello_repositories_use_koji=True|False```
- ```foreman_repositories_use_koji=True|False```

have been renamed to:
- ```katello_repositories_environment=staging|release```
- ```foreman_repositories_environment=staging|release```

where ```staging``` => koji, ```release``` => nightly

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1743,7 +1743,7 @@ def apply_hotfix():
 
 def upstream_install(admin_password=None, run_katello_installer=True):
     """Task to install Foreman nightly using forklift scripts"""
-    use_koji = 'koji' in os.environ.get('DISTRIBUTION').lower()
+    koji = 'koji' in os.environ.get('DISTRIBUTION').lower()
     puppet4 = bool(os.environ.get('PUPPET4_REPO'))  # otherwise puppet5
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1763,8 +1763,8 @@ def upstream_install(admin_password=None, run_katello_installer=True):
             '-e foreman_installer_skip_installer=True '
             'playbooks/katello.yml'.format(
                 '-e puppet_repositories_version=4' if puppet4 else '',
-                '-e foreman_repositories_use_koji=True' if use_koji else '',
-                '-e katello_repositories_use_koji=True' if use_koji else '',
+                '-e foreman_repositories_environment=staging' if koji else '',
+                '-e katello_repositories_environment=staging' if koji else '',
```
kbidarkar: I did not get a chance to test this out, so is it that by default, it is `nightly/release` if `foreman_repositories_environment=` is not specified?

lpramuk: implicit default is 'release' so no need to be explicit about it

kbidarkar: ok, just wanted to confirm.

PR #667, opened 2018-03-19
Handle epel installs/removals differently

lpramuk:
- the logic is if epel was enabled before an action leave it also enabled afterwards (useful for nightly)
- otherwise if epel has been enabled then remove it afterwards (usual downstream)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -237,10 +237,13 @@ def setup_avahi_discovery():
        by 'ping vm.local' run at Satellite
     """
     os_version = distro_info()[1]
-    run('yum -y install https://dl.fedoraproject.org/pub/epel/'
-        'epel-release-latest-{0}.noarch.rpm'.format(os_version))
+    epel_present = run('rpm -q epel-release', warn_only=True).return_code == 0
+    if not epel_present:
+        run('yum -y install https://dl.fedoraproject.org/pub/epel/'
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -237,10 +237,13 @@ def setup_avahi_discovery():
        by 'ping vm.local' run at Satellite
     """
     os_version = distro_info()[1]
-    run('yum -y install https://dl.fedoraproject.org/pub/epel/'
-        'epel-release-latest-{0}.noarch.rpm'.format(os_version))
+    epel_was_missing = not run('rpm -qa epel-release')
```
lpramuk: Note: rpm -qa epel-release
$?=0 always

lpramuk: so no need for warn_only=True

kbidarkar: How about we do it this way?
```
epel_missing = run('rpm -qa epel-release')
if "epel-release" not in epel_missing:
     run('yum -y install https://dl.fedoraproject.org/pub/epel/'
           'epel-release-latest-{0}.noarch.rpm'.format(os_version))
```

lpramuk: I've just realized that fabric supports ```return_code```
Using ```return_code``` it would be most precise.

lpramuk:
Comments resolved.

PR #672, opened 2018-03-22
Using six library for both python2 and python3 compatibility

kbidarkar:
This would help achieve importing the compatible libs with less code.

Signed-off-by: Kedar Bidarkar <kbidarka@redhat.com>


kbidarkar:
Had thought of using the six module, but then later realized we have `future` module, which is recommended over `six` module, unless we want some support for very older Python versions.

kbidarkar:
@ldjebran mentioned we are already using `six` module in robottelo, so will make use of `six` module for this.

PR #674, opened 2018-03-25
No python-rhsm on RHEL7.5

lpramuk:
python-rhsm was renamed to subscription-manager-rhsm...

Anyways, the check for ```python-rhsm``` is pretty useless since ```subscription-manager``` requires it always.
So checking for subscription-manager presence is sufficient and implies python-rhsm presence.

RHEL7.5
```
# rpm -qR subscription-manager | grep rhsm
subscription-manager-rhsm = 1.20.10
```
RHEL7.4
```
# rpm -qR subscription-manager | grep rhsm
python-rhsm >= 1.19.10
```


kbidarkar:
ACK

PR #675, opened 2018-03-29
Remove sat6.3 to be run as INTERNAL, when we have INTERNAL_AK

kbidarkar:
Signed-off-by: Kedar Bidarkar <kbidarka@redhat.com>

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2065,8 +2065,9 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     }
     distribution = distribution.lower()
 
-    # Make sure downstream_install only is called for sat6.1, sat6.2, sat6.3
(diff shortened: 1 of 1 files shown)


JacobCallahan:
Would this mean I can't provide an internal compose url to the installer job, if I'm trying to install 6.3?

omaciel:
@JacobCallahan I believe this change only affects jobs using the `INTERNAL AK` parameter

lpramuk:
This means that for 6.3 default is INTERNAL_AK

lpramuk:
Merged-in-by: Lukas Pramuk lpramuk@redhat.com

kbidarkar:
This means,
1) `INTERNAL`, `INTERNAL AK` and `INTERNAL REPOFILE` would work from sat6.3+
2) ONLY `INTERNAL` for Sat6.1 and Sat6.2, not the other 2 options.

PR #678, opened 2018-04-10
Fix avahi error on RHEL 7.5

JacobCallahan:
This change simply restarts the dbus which seems to fix the issues
we've been seeing on RHEL 7.5.
This fix does not affect RHEL 6.

Fixes #642

Issue #642, opened 2017-12-20
avahi not a friend with rhel7.5


pondrejk:
Just FYI, with vault req. testing I found avahi as installed in automation job fails to start on rhel7.5 (job no 3602). Going to work around this by running automation from a separate branch, but filing this issue as it can cause trouble in the future.

JacobCallahan:
from my jobs
```
[mybox] run: service avahi-daemon restart
[mybox] out: Redirecting to /bin/systemctl restart avahi-daemon.service
[mybox] out: Job for avahi-daemon.service failed because the control process exited with error code. See "systemctl status avahi-daemon.service" and "journalctl -xe" for details.
[mybox] out: 

Fatal error: run() received nonzero return code 1 while executing!
```

from my satellite box
```
-bash-4.2# systemctl status avahi-daemon.service
● avahi-daemon.service - Avahi mDNS/DNS-SD Stack
   Loaded: loaded (/usr/lib/systemd/system/avahi-daemon.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Tue 2018-04-10 20:56:40 CEST; 7min ago
  Process: 29157 ExecStart=/usr/sbin/avahi-daemon -s (code=exited, status=255)
 Main PID: 29157 (code=exited, status=255)
   Status: "avahi-daemon 0.6.31 exiting."

Apr 10 20:56:40 mybox avahi-daemon[29157]: Found user 'avahi' (UID 70) and group 'avahi' (GID 70).
Apr 10 20:56:40 mybox avahi-daemon[29157]: Successfully dropped root privileges.
Apr 10 20:56:40 mybox avahi-daemon[29157]: avahi-daemon 0.6.31 starting up.
Apr 10 20:56:40 mybox avahi-daemon[29157]: dbus_bus_request_name(): Connection ":1.90" is not allowed to own the service "org.freedesktop.Avahi" d...tion file
Apr 10 20:56:40 mybox avahi-daemon[29157]: WARNING: Failed to contact D-Bus daemon.
Apr 10 20:56:40 mybox avahi-daemon[29157]: avahi-daemon 0.6.31 exiting.
Apr 10 20:56:40 mybox systemd[1]: avahi-daemon.service: main process exited, code=exited, status=255/n/a
Apr 10 20:56:40 mybox systemd[1]: Failed to start Avahi mDNS/DNS-SD Stack.
[...]

PR #689, opened 2018-04-17
Hands off of dbus

lpramuk:
Revert "Fix avahi error on RHEL 7.5" (PR https://github.com/SatelliteQE/automation-tools/pull/678)

By restarting dbus we've run into cascading effect of what services to restart.
DBus shall not be restarted, see systemd/systemd#2925 (comment)

After dbus restart every ssh connection receives 25sec penalty. Then you need to restart systemd-logind.service to wear off this penalty, so I'm not gonna hunt what else to restart rather I'm reverting and will investigate the original issue with avahi once again.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -243,7 +243,6 @@ def setup_avahi_discovery():
     if os_version >= 7:
         run('firewall-cmd --add-service mdns --permanent')
         run('firewall-cmd --reload')
-        run('service dbus restart')
     else:
         run('iptables -I INPUT -d 224.0.0.251/32 -p udp -m udp --dport 5353'
             ' -m conntrack --ctstate NEW -j ACCEPT')


lpramuk:
Checked upstream provisioning jobs:
77min  vs.  48min  (before)

You wonder about extra 30mins right? It's because we restarted dbus !!!

JacobCallahan:
@lpramuk @san7ket this revert did indeed reintroduce #642 in both of the installer jobs I ran this morning. Is there a way to fix this **without** specifically restarting dbus?

lpramuk:
What environment do you use? 
We didn't hit that upon retesting. Both in VLAN and beaker

lpramuk:
Ah I see now 
@JacobCallahan with RHEL7.4 you need to use avahi 0.6.31-17.el7 but you are using avahi released for RHEL7.5  - In other words upgrading 7.4 to 7.5 without rebooting

We don't have to go adding reboot( or various service restarts) if we just use latest RHEL7.5 machines 
Fresh and updated (with reboot) is explicit requirement. We can directly ask beaker to give us such machine. So please start using them too

No need to test OS upgrade before Satellite bits touched the machine. 
In place upgrade that something different - worth to implement (it would involve reboot)

JacobCallahan:
@lpramuk Using the latest and updated version of RHEL is not an explicit requirement for our installer job. This is why the installer performs and update before attempting to install Satellite. Also, beaker is too slow to update their available OS options to provide a RHEL 7.5 box on the day it is released. It was during our required RHEL 7.5 GA install test that we first encountered this issue.

So the installer job should continue to perform updates to keep the target systems as up-to-date as possible. It should also be able to handle rebooting a system when necessary (likely whenever a kernel update is applied).  Avahi is just the latest issue encountered because we don't reboot, there are likely more to come in the future and we should avoid them as best we can.

How to go about deciding when to reboot and how to handle the reboot is something the team can discuss and come up with the best solution. However, it is my strong opinion that this is something we absolutely need to do.
@SatelliteQE/quality-engineers

PR #703, opened 2018-04-30
setup_foreman_discovery - template changed since 6.4

rplevka:
since 6.4, the default template no longer uses hardcoded ONTIMEOUT parameter, but rather has it parametrized:
```
ONTIMEOUT <%= global_setting("default_pxe_item_global", "local") %>
```

this makes the original `sed` command miss the line:
```
sed -i -e "s/^ONTIMEOUT\s\+local/ONTIMEOUT discovery/"
```

the solution would be to update the regexp to match the pattern again, however the neater solution is to use the proper `hammer settings set` command to upload the default parameter.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1055,8 +1055,14 @@ def setup_foreman_discovery(sat_version):
     run('hammer -u admin -p {0} template dump --name '
         '"PXELinux global default" > {1}'.format(
             admin_password, template_file))
-    run('sed -i -e "s/^ONTIMEOUT\s\+local/ONTIMEOUT discovery/" {0}'
-        .format(template_file))
+    if sat_version in ('6.1', '6.2', '6.3'):
+        # since 6.4, ONTIMEOUT option uses "default_pxe_item_global" setting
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1055,8 +1055,13 @@ def setup_foreman_discovery(sat_version):
...
         '"PXELinux global default" > {1}'.format(
             admin_password, template_file))
-    run('sed -i -e "s/^ONTIMEOUT\s\+local/ONTIMEOUT discovery/" {0}'
-        .format(template_file))
+    if sat_version == '6.3':
+        # since 6.4, ONTIMEOUT option uses "default_pxe_item_global" setting
+        run('sed -i -e "s/^ONTIMEOUT\s\+local/ONTIMEOUT discovery/" {0}'
+            .format(template_file))
+    else:
+        run('hammer -u admin -p {0} settings set --name '
```
ntkathole: @rplevka value of {0} ?

rplevka: missed that during copy&pasta..great catch

rplevka: fixed. pls re-review

PR #705, opened 2018-05-17
Refactor susbscribe() so it handles cleanup and stage registration

### LINKED ISSUE #443: Make switching between AK to REPOFILE install method possible
If for some reason we may want to move from AK to REPOFILE install method, 
one needs to run the below commands.

1) subscription-manager unregister
2) subscription-manager clean 
3) fab -H root@fqdn-hostname clean_rhsm
4) Remove any *.repo files in /etc/yum.repos.d/
5) yum clean all
6) Rebuild using REPOFILE

lpramuk:
Upon re-running installation if you switch distribution from "INTERNAL AK"  (to INTERNAL or CDN) then registration fails

Fixes #443

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -36,7 +36,7 @@ def unsubscribe():
     run('subscription-manager clean')
 
 
-def subscribe(autosubscribe=False):
+def subscribe(autosubscribe=False, stage=False):
     """Registers and subscribes machine to Red Hat.
 
     The following environment variables affect this command:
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2216,16 +2226,12 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     # When creating a vm the vm_ip will be set, otherwise use the fabric host
     host = env.get('vm_ip', env['host'])
 
-    # Register and subscribe machine to Red Hat
-    if distribution == 'satellite6-cdn' and test_in_stage:
-        execute(update_rhsm_stage, host=host)
-
```
lpramuk: This is now handled within subscribe()

PR #712, opened 2018-06-08
Avoid epel and install nss-mdns directly since it is just 1 rpm (with deps in rhel)

lpramuk:
- rather update epel7 ```nss-mdsn``` package version from time to time than have EPEL failing any time (epel6 nss-mdsn rpm is stalled for over 8 years)


kbidarkar:
@lpramuk, feel we avoid this. :)

kbidarkar:
Had a discussion with @lpramuk, we agreed to close this PR.

PR #718, opened 2018-06-20
Enable foreman-maintain repo for satellite 6.4 installation

ntkathole:
https://bugzilla.redhat.com/show_bug.cgi?id=1592023
Satellite 6.4 installation requires foreman-maintain as dependency

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2300,6 +2300,18 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         puppet4_repo = os.environ.get('PUPPET4_REPO')
         if puppet4_repo:
             execute(create_custom_repos, puppet4_repo=puppet4_repo, host=host)
+    # Sat6.4 and above: enable internal foreman-maintain repo for installation
+    if (
+            sat_version not in ['6.1', '6.2', '6.3']
+            and not distribution.endswith('cdn')
+    ):
+        maintain_repo = os.environ.get('MAINTAIN_REPO')
+        if maintain_repo:
+            execute(
+                create_custom_repos,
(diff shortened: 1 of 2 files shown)


lpramuk:
Please fix flake8 issues first

automation_tools/repository.py
```diff
@@ -153,6 +153,7 @@ def enable_satellite_repos(cdn=False, beta=False, disable_enabled=True,
     ]
     if sat_version == '6.4':
         repos.append('rhel-{0}-server-ansible-2-rpms')
+        repos.append('rhel-7-server-satellite-maintenance-6-rpms')
```
lpramuk: could be {0} instead of '7' 

if future then future, you considered sat 6.5,6.6... but not rhel8 :-)

ntkathole: no it will be only 6 no matter what sat version is...currently it has support for all 6.1, 6.2, 6.3 and 6.4.

automation_tools/__init__.py
```diff
@@ -2300,6 +2300,13 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         puppet4_repo = os.environ.get('PUPPET4_REPO')
         if puppet4_repo:
             execute(create_custom_repos, puppet4_repo=puppet4_repo, host=host)
+    # Sat6.4 and above: enable internal foreman-maintain repo for installation
+    if sat_version not in ['6.1', '6.2', '6.3'] and not distribution.endswith('cdn'):# noqa
```
kbidarkar: Better we drop  "#noqa" here.

ntkathole:
@kbidarkar updated

PR #723, opened 2018-08-10
Update kickstart templates

lpramuk:
- fixes VM timezone settings when OS sets `America/New_York` while HW clock is set to `UTC`
- fixes unreliable mDNS discovery when record disappears after 2mins
- do as much as possible in postinstall time since network is set up (ssh keygen, authorized keys)
- fixes authorized keys put in rhel6 postinstall has wrong selinux fcontext
- rc.local logs both on console and in the log
- handle IPv6 disablement


lpramuk:
No, rhel610 is already done like that, rhel76beta too, I would wait for rhel76 GA, no re-creating

PR #727, opened 2018-09-03
Remove the sleep and replace with host ssh check method in automation

san7ket:


omaciel:
The body for both new methods share a lot of similar code and logic... How about refactoring it so that the method for checking if a host is up (i.e. can be 'pinged') and connected to (i.e. can be 'ssh'ed into') use the same code and then the different methods can be simplified into only a few lines of code? This also has the potential of allow us to use the same logic to check for other things.

Just a suggestion :)

ntkathole:
Merging as need to fix failure in trigger

PR #735, opened 2018-10-24
Automatic baseimage creation and deployment

lpramuk:
Two new tasks to be imployed by future jenkins job: 
- ```create_baseimage()```
- ```deploy_baseimage()```

### DIFF
--- a/automation_tools/baseimage.py
+++ b/automation_tools/baseimage.py
@@ -1,0 +1,133 @@
+"""Tasks for base image creation and deployment"""
+from __future__ import print_function
+
+import os
+import time
+
+from fabric.api import env, get, put, run, settings
+from re import search, MULTILINE
+
+from six.moves.urllib.parse import urljoin
+
+
+def detect_imagename(os_url):
+    """Task to detect image name by OS URL
+
+    :param str os_url: URL of OS media to detect
+
+    """
+    comp_id = run('wget -q -O- {}'.format(urljoin(os_url, '../../../../COMPOSE_ID')), quiet=True)
+    if comp_id.succeeded:
+        match_comp = search(r'(\w+)-([\d\.]+)-(?:\w+-)?([\d\.]+)', comp_id)
+        image = match_comp.group(1).lower() + match_comp.group(2).replace('.', '') + '-' + \
+            match_comp.group(3)
+    else:
+        image = 'unknown-{}'.format(str(time.time()).split('.')[0])
+
+    print(image)
+    return(image)
+
+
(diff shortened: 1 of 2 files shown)


ntkathole:
@lpramuk If read the code correctly, the created base image wont have DNS entry of IDM server. Can we have step to update /etc/resolv.conf in `create_baseimage()` function else we need to again update image manually? Also same for /etc/ntp.conf. Let me know if I am missing something here.

automation_tools/baseimage.py
```diff
@@ -0,0 +1,98 @@
...
+                match_comp.group(3)
+        else:
+            image = 'unknown-{}'.format(str(time.time()).split('.')[0])
+
+    put('misc/base_image_creation/ks_rhel{}_template'.format(os_ver), 'ks.cfg')
+    run('sed -i "s|OS_URL|{}|g" ks.cfg'.format(os_url))
+    run('sed -i "s|ENCRYPT_PASS|$1$xyz$7xHVh4/yhE6P00NIXbWZA/|g" ks.cfg')
+    run('sed -i "s|AUTH_KEYS_URL|{}|g" ks.cfg'.format(os.getenv('AUTH_KEYS_URL')))
+
+    run('virt-install --connect qemu:///system -n {img} -l {url} -w bridge:br0 '
```
rplevka: do we need to use the `bridge:br0` here? if one would liek to run this on his local or cusotm libvirt, he probalby won't have such bridge configured. + i don't think we need it there. 

not a blocker

rplevka: don't we want to include the qemu-guest-agent serial interface here? or shall we leave this on snap-guest?

rplevka:
@ntkathole - I would not hardcode the environment-specific configuration. Perhaps we might make the task to accept parameters tho

lpramuk:
Yeah I know about DNS, I thought it would be Step2... But I changed my mind and incorporated in PR
DNS parametrization needs change in ks templates to works, but won't fail now.

So, changes: 
- added more parametrization of base image
- create all-in-one func ```deploy_baseimage_by _url()``` combining create and deploy together as in shell you cannot consume func return
- detect baseimage name code moved to separate func ```detect_imagename()```

automation_tools/baseimage.py
lpramuk: The latter. It has to be left for snap-guest, here we are creating only baseimage that has to have qemu-guest-agent installed (we do it already). Therefore serial interface has to added to cloned VMs and not to baseimage VM. (but could be for nothing, no loss, no gain)

PR #737, opened 2018-10-29
Fix capsule-certs-generate task and discovery setup

ntkathole:


automation_tools/__init__.py
```diff
@@ -1052,7 +1052,7 @@ def setup_foreman_discovery(sat_version):
         run('wget -nv -O- {0} | tar x --overwrite -C /var/lib/tftpboot/boot'
             .format(image_url))
 
-    if sat_version in ('6.3', '6.4', 'downstream-nightly'):
+    if sat_version in ('6.3', '6.4', '6.5', 'downstream-nightly'):
```
san7ket: Can we reverse the logic here please, non blocking comment

ntkathole:
Need it to trigger auto...Merging!

PR #739, opened 2018-10-30
fix partition_disk task

### LINKED ISSUE #738: trouble with partition_disk task
Installer with partition disk task enabled fails with:

```
sed -i '|/home|d' /etc/fstab
sed: -e expression #1, char 1: unknown command: `|'

Fatal error: run() received nonzero return code 1 while executing!
```

Introduced in #736

pondrejk:
fixes #738


lpramuk:
pls no noqa, you can escape it twice to avoid W605:
```
run("sed -i '/\\/home/d' /etc/fstab")
```

PR #751, opened 2019-01-29
disable_base_os checks also os_upgrade_repos

### LINKED ISSUE #747: trouble with disable_base_os in provisioning jobs
When attempting to run a provisioning job with custom rhel image (in my case needed for fips testing) the logic introduced in commits https://github.com/SatelliteQE/automation-tools/pull/593 and https://github.com/SatelliteQE/automation-tools/pull/594 disables the base os repo, which prevents installation from completing. 

I don't know why such logic was introduced, but we still should be able to have the case with custom source image **and** default os repo enabled. IMHO disable_base_os should be only controlled by what is set in OS_UPGRADE_REPOS

pondrejk:
fixes #747, helps running at with custom fips image that requires base-os for yum update afterwards.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1134,7 +1134,7 @@ def disable_baseos_repo():
     IMAGE
         The custom image name to be used for vault_requests.
     """
-    if os.environ.get('IMAGE'):
+    if os.environ.get('OS_UPGRADE_REPOS'):
         os_version = distro_info()[1]
         disable_repos('rhel-{0}-server-rpms'.format(os_version))
 


san7ket:
Wouldn't this cause issue when we need to check vault request

pondrejk:
@san7ket I don't think so, because iirc in that case we would provide the OS_UPGRADE_REPOS with latest bits for the os, so the condition would be met

pondrejk:
@lpramuk using "or" wouldn't help me in the situation when there is IMAGE set but not the OS_UPGRADE_REPOS, in such case I don't want to disable rhel-7-server-rpms. On the second look, i think it could be just 

`if os.environ.get('OS_UPGRADE_REPOS'):`

without regard to custom image. So if someone wants to have custom repos, he needs to set it and not expect to have it present in the image. Does it make sense?

PR #755, opened 2019-02-28
Add RHEL8 ks template

lpramuk:
Add RHEL8 ks template for baseimage creation

(btw Fix E117 over-indented)

PR #758, opened 2019-03-07
Update AppStream URL in rhel8_ks_template

lpramuk:
I missed these changes to send in along with PR #755 

- task ```create_baseimage()``` now updates AppStream URL in rhel8_ks_template
- script ```create-base-image.sh``` now supports RHEL8

### DIFF
--- a/automation_tools/baseimage.py
+++ b/automation_tools/baseimage.py
@@ -50,13 +50,15 @@ def create_baseimage(os_url, image=None, auth_keys_url=None, dns_server=None, di
         if match_name:
             os_ver = match_name.group(2).split('.')[0]
         else:
-            os_ver = 7
+            os_ver = 8
 
     if not image:
         image = detect_imagename(os_url)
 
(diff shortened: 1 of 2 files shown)


automation_tools/baseimage.py
```diff
@@ -57,6 +57,7 @@ def create_baseimage(os_url, image=None, auth_keys_url=None, dns_server=None, di
 
     put('misc/base_image_creation/ks_rhel{}_template'.format(os_ver), 'ks.cfg')
     run('sed -i "s|OS_URL|{}|g" ks.cfg'.format(os_url))
+    run('sed -i "s|AS_URL|{}|g" ks.cfg'.format(os_url.replace('BaseOS', 'AppStream')))
```
san7ket: This should be run for just rhel8 right?

lpramuk: Yes for RHEL8

For RHEL7 it's safe since ks template has no "AS_URL" (sed replace nothing)

lpramuk: I changed my mind and incorporated change into PR

PR #761, opened 2019-04-09
add setup_bfa_prevention to set the failed login attempts limit

rplevka:
since 6.5, there's a global setting allowing us to control the Brute-force-attack prevention mechanism by specifying the failed-attempt count limit.
This task allows us to control it from `product_install` task.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -516,6 +516,26 @@ def setup_default_subnet(sat_version):
         run(command.replace(' create ', ' update ', 1))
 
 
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -516,6 +516,26 @@ def setup_default_subnet(sat_version):
         run(command.replace(' create ', ' update ', 1))
 
 
+def setup_bfa_prevention(sat_version):
```
san7ket: +1

Issue #762, opened 2019-04-12
Fix task setup_foreman_discovery for nightlies


san7ket:
```
run: rm -f /etc/yum.repos.d/centos_atomic.repo
Executing task 'setup_foreman_discovery'
 run: wget -nv -O- http://downloads.theforeman.org/discovery/nightly/fdi-image-latest.tar | tar x --overwrite -C /var/lib/tftpboot/boot
] out: 2019-04-11 17:15:00 URL:http://downloads.theforeman.org/discovery/nightly/fdi-image-latest.tar [218798080/218798080] -> "-" [1]
 out: 

rhel 7 6
 run: systemctl restart foreman
 out: Failed to restart foreman.service: Unit not found.
 out: 
```

ntkathole:
This looks something change from foreman side or it is a new issue that service not available.
Earlier we have foreman daemon on this system but now it shows Unit not found.

https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L1080

PR #763, opened 2019-04-13
Restarting foreman services in setup_discovery not needed

ntkathole:
Restarting services not needed in setup_discovery task.
Now in upstream nightly, foreman service is not available (t is part of foreman-service rpm which is not mandatory.). so job is failing. 
Refs : foreman-packaging/pull/3315


ntkathole:
Fixes #762

PR #768, opened 2019-05-10
disable usage of dns by sshd

rplevka:
the functionality is not critical and causes big time delays in case the nameservers are not reachable


rplevka:
ah, hang on - apparently we're operating on `ks.cfg`. my solution won't work. Looking for a solution.

rplevka:
repoened as https://github.com/SatelliteQE/automation-tools/pull/769

PR #769, opened 2019-05-10
disable usage of dns by sshd

rplevka:
the functionality is not critical and causes big time delays in case the nameservers are not reachable


san7ket:
@rplevka bump?

rplevka:
@lpramuk @san7ket - comments addressed

PR #773, opened 2019-05-13
Fix temporary RHEL8 avahi install until EPEL8 is out

lpramuk:
- install avahi (nss-mdns) only if not present (faster for re-runs)
- rhel8 temporary fix until EPEL8 with new nss-mdns is out

Works around #774


san7ket:
Can you also create a issue for this, so we track it

lpramuk:
#774

PR #781, opened 2019-05-28
Clean up 6.1

lpramuk:
- dropped 6.1 support
- avoided to clean every occurrence of ```6.1``` string, sometimes was useful to keep ```6.1``` string to maintain original logic of the clause

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -373,41 +373,36 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
             print('Was not possible to fetch interface information')
             sys.exit(1)
 
-    if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1'):
-        proxy = 'capsule'
-    else:
-        proxy = 'foreman-proxy'
-
     installer_options = {
-        '{0}-dns'.format(proxy): 'true',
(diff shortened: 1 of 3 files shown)


automation_tools/__init__.py
```diff
@@ -373,41 +373,36 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
...
+        'foreman-proxy-tftp': 'true',
+        'foreman-proxy-tftp-servername': hostname,
+        'foreman-proxy-register-in-foreman': 'true',
     }
 
     # Handles enabling puppet support
     if not os.environ.get('SATELLITE_VERSION') in ('6.3', '6.4'):
-        installer_options['{0}-puppetca'.format(proxy)] = 'true'
+        installer_options['foreman-proxy-puppetca'] = 'true'
         if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1', '6.2'):
```
rplevka: isn't it safe to reduce this to `6.2`?

lpramuk: ok, wanted to maintain original context as it seems we miss to add ```'6.5'``` and ```'6.6'``` to ```('6.3', '6.4')```

PR #783, opened 2019-05-29
Clean up 6.2

lpramuk:
- dropped 6.2 support
- refactored ```setup_bfa_prevention()``` task to be run conditionally

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -387,14 +387,6 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
         'foreman-proxy-register-in-foreman': 'true',
     }
 
-    # Handles enabling puppet support
-    if not os.environ.get('SATELLITE_VERSION') in ('6.3', '6.4'):
-        installer_options['foreman-proxy-puppetca'] = 'true'
-        if os.environ.get('SATELLITE_VERSION') in ('6.0', '6.1', '6.2'):
-            installer_options['capsule-puppet'] = 'true'
-        else:
-            installer_options['foreman-proxy-puppet'] = 'true'
-
     installer_options[
         'foreman-proxy-dhcp-range'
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -373,41 +373,28 @@ def setup_default_capsule(interface=None, run_katello_installer=True):
...
+        'foreman-proxy-dns-zone': domain,
+        'foreman-proxy-dhcp': 'true',
+        'foreman-proxy-dhcp-interface': interface,
+        'foreman-proxy-tftp': 'true',
+        'foreman-proxy-tftp-servername': hostname,
+        'foreman-proxy-register-in-foreman': 'true',
     }
 
-    # Handles enabling puppet support
-    if not os.environ.get('SATELLITE_VERSION') in ('6.3', '6.4'):
```
lpramuk: Here we missed to add newer versions (6.5, 6.6)
So when considering removal of this code read this as 
```
if not os.environ.get('SATELLITE_VERSION') in ('6.3', '6.4', '6.5', '6.6'):
```

lpramuk: Anyway this options aren't needed for 6.3 and older since they are both true by default (tested)

automation_tools/__init__.py
```diff
@@ -512,32 +499,28 @@ def setup_default_subnet(sat_version):
...
 
-def setup_bfa_prevention(sat_version):
+def setup_bfa_prevention(bfa_limit=0):
     """Postinstall task to set the `failed_login_attempts_limit` global setting
     to control brute-force-attack prevention.
 
     Expects the following environment variables:
 
-    BFA_LIMIT
-        number of failed login attempts in 5 minute interval to trigger BFA prevention
```
lpramuk: env var not used, let it better be a param set to 0 by  default

lpramuk:
Tested many times as I'm have my other PRs based of this branch and running tests daily when developing the other PRs

PR #787, opened 2019-06-25
Implement possibility to validate manifest

jhutar:
Fixes https://github.com/SatelliteQE/automation-tools/issues/786

This is how config file would look like:

```
$ cat config/robottelo-manifest-content.conf 
# This file contains subscription names (one per line) expected in robottelo-automation manifest.
# Subscriptions named here must be in the manifest and no other subs are allowed.
Red Hat Beta Access
Red Hat Enterprise Linux for Virtual Datacenters, Premium
Red Hat Enterprise Linux Server, Premium (8 sockets) (Unlimited guests)
Red Hat Enterprise Linux Server, Premium (Physical or Virtual Nodes)
Red Hat Satellite Infrastructure Subscription
```

PR #788, opened 2019-06-26
Adding assert before downloading manifest

schlupov:


automation_tools/__init__.py
```diff
@@ -3115,13 +3116,39 @@ def download_manifest(url=None, consumer=None):
                          ' session and distributor hash')
 
 
+def assert_downloading_manifest():
```
rplevka: i'd call this `verify_attached_subscriptions` or something similar.

automation_tools/__init__.py
```diff
@@ -3115,13 +3116,39 @@ def download_manifest(url=None, consumer=None):
...
+    """Makes sure that downloaded manifest matches manifest name from a reference file."""
```
rplevka: This is not really true statement, as we don't check a downloaded manifest - this method checks that the portal account has the required subscriptions attached

automation_tools/__init__.py
```diff
@@ -3115,13 +3116,39 @@ def download_manifest(url=None, consumer=None):
...
+
+    user = os.environ.get('RHN_USERNAME')
+    password = os.environ.get('RHN_PASSWORD')
+    consumer = os.environ.get('CONSUMER')
+    product_names_file = os.environ.get('PRODUCT_NAMES')
+
+    r = requests.get('https://subscription.rhn.redhat.com/subscription/'
+                     'consumers/{0}/entitlements'.format(consumer),
+                     auth=(user, password), verify=False)
```
rplevka: How about adding a
`r.raise_for_status()` line here, so the script raises a proper HTTP exception in case the server does not return 200?

automation_tools/__init__.py
```diff
@@ -3115,13 +3116,39 @@ def download_manifest(url=None, consumer=None):
...
+    rhn = r.json()
+
+    product_names = open(product_names_file, 'r')
+    line = product_names.readline()
+    product_counter = 0
+    while line:
+        subscription = rhn[product_counter]["pool"]["productName"]
+        assert line != subscription, "Product name {0} doesn't match {1} from file!".format(
```
rplevka: if this fails - doesn't it mean it keeps the file opened?
I would wrap the whole block inside a context manager, that assures the file is properly closed in any case..
```
with open(product_names_file, 'r'):
   the rest of code
```

rplevka: how about collecting all the avalable subs into a list by using list comprehension and then simply iterate?
```python
with open(product_names_file, 'r') as subs:
    for sub in subs:
        assert sub in [s[product_counter]["pool"]["productName"] for s in rhn], "Subscription for product {0} is not present".format(sub)
```

automation_tools/__init__.py
```diff
@@ -3115,13 +3116,53 @@ def download_manifest(url=None, consumer=None):
...
+    with open(product_names_file, 'r') as product_names:
+        line = product_names.readline()
+        product_counter = 0
+        while line:
+            try:
+                subscription = rhn[product_counter]["pool"]["productName"]
+                if line.rstrip() != subscription.rstrip():
+                    missing_product.append(subscription.rstrip())
+            except IndexError:
+                missing_product.append(line.rstrip())
```
jhutar: Is this:
```
try:
  ...
  missing_product.append(...)
except ...:
  missing_product.append(...)
```
correct?

jhutar:
Closing as this duplicates https://github.com/SatelliteQE/automation-tools/pull/787 I think. Thank you!

PR #789, opened 2019-07-07
Remove enable_ostree() task and add installer option instead

lpramuk:
This fixes upstream where running installer with ```--katello-enable-ostree=true``` additionally causes pulp misbehavior while running this option at once causes no issues to pulp.

So it will speed up any SAT installation saving one (redundant) installer run

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -16,8 +16,7 @@ from re import search
 
 from automation_tools.bz import bz_bug_is_open
 from automation_tools.repository import (
-    create_custom_repos, delete_custom_repos, disable_repos,
-    disable_beaker_repos, enable_repos, enable_satellite_repos,
+    create_custom_repos, disable_repos, disable_beaker_repos, enable_repos, enable_satellite_repos,
 )
 from automation_tools.utils import (
     distro_info, run_command, update_packages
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -1823,16 +1799,15 @@ def apply_hotfix():
 
 def upstream_install(admin_password=None, run_katello_installer=True):
     """Task to install Foreman nightly using forklift scripts"""
-    koji = 'koji' in os.environ.get('DISTRIBUTION').lower()
+    koji = 'koji' in os.environ.get('SATELLITE_DISTRIBUTION').lower()
```
ntkathole: We are exporting SATELLITE_DISTRIBUTION in installer and other jobs, are you going to handle this in any PR?

lpramuk: Yeah, so I'm fixing it ```s/DISTRIBUTION/SATELLITE_DISTRIBUTION/```

@ntkathole you seem to me you misread the direction (+/- lines)

lpramuk: Thanks you made me think and I improved the code to:
```
koji = 'koji' in os.environ.get('SATELLITE_DISTRIBUTION','').lower()
```
```''.lower()``` is much better than ```None.lower()``` 
Now it will work even if there is no ```SATELLITE_DISTRIBUTION``` defined ;-)

ntkathole: @lpramuk we have DISTRIBUTION exported in satellite-installer job but not SATELLITE_DISTRIBUTION

lpramuk: @ntkathole you are right we have spaghetti code in ```satellite6-provisioning.sh```

https://github.com/SatelliteQE/robottelo-ci/blob/master/scripts/satellite6-provisioning.sh#L33-L49

PR #791, opened 2019-07-20
Fix capsule-cert-generate command

ntkathole:
Fixed job/capsule-sanity-check-6.6-rhel7/6/console

Regex changed due to https://bugzilla.redhat.com/show_bug.cgi?id=1709761


ntkathole:
@san7ket @lpramuk Needed this to fix capsule sanity job. Can you guys look into it?

PR #792, opened 2019-07-31
Added support to attach missing subscription if missing

vijay8451:
- Added support to attach missing subscription if missing.
- MR:  245 to add pool_id and count under `robottelo-manifest-content.conf `

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -3034,9 +3034,8 @@ def setenforce(mode):
     run('setenforce {0}'.format(mode))
 
 
-def download_manifest(url=None, consumer=None, pool_id=None, count=None):
-    """Task for downloading the manifest file from customer portal,
-    also attach subscription to it if requires.
+def download_manifest(url=None, consumer=None):
+    """Task for downloading the manifest file from customer portal.
 
     The following environment variables affect this command:
 
@@ -3051,8 +3050,6 @@ def download_manifest(url=None, consumer=None, pool_id=None, count=None):
 
     :param url: Subscription Manager URL
     :param consumer: A consumer hash to be used for getting the manifest
-    :param pool_id: A pool ID of subscription
-    :param count: A quantity of subscription requires to attach
     :returns: a path string to a downloaded manifest file
     """
     user = os.environ.get('RHN_USERNAME')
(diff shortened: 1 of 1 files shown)


ntkathole:
Tested using jobs /job/satellite6-manifest-downloader/2240/. Works perfect.
Only enhancement can be, we should be able to delete extra subscriptions added so that manifest size should not increase if it is touched by someone. We can deal with it in separate PR.

vijay8451:
ping @jhutar

automation_tools/__init__.py
```diff
@@ -3067,61 +3070,84 @@ def download_manifest(url=None, consumer=None):
...
+            continue
+        else:
+            response.append('Yes')
+            print("Attaching missing subscription %s" % (sub))
+            attach = download_manifest(pool_id=exp_subs[sub][0], count=exp_subs[sub][1])
+            if attach:
+                print("Successfully attached subscription %s" % (sub))
+            else:
+                return [False]
+    return response
```
jhutar: I dislike this complex undocumented return value of the function. If we would have separate function to fix entitlements, this could stay very simple. Or at least return touple instead of list?

automation_tools/__init__.py
```diff
@@ -3133,7 +3159,10 @@ def relink_manifest(manifest_file=None):
     if manifest_file is None:
         manifest_file = download_manifest()
         if os.environ.get('EXP_SUBS_FILE', None) is not None:
-            assert validate_manifest(manifest_file)
+            validate = validate_manifest(manifest_file)
+            assert validate[0]
+            if len(validate) > 1:
+                manifest_file = download_manifest()
```
jhutar: I think that this means we just re-download the manifest if its validation fails. In robottelo-ci for download job we are calling `relink_manifest`:

    https://github.com/SatelliteQE/robottelo-ci/blob/master/scripts/satellite6-manifest-downloader.sh

Shall we rerun validation after second `download_manifest` to be sure all is fine now?

automation_tools/__init__.py
```diff
@@ -3034,8 +3034,9 @@ def setenforce(mode):
     run('setenforce {0}'.format(mode))
 
 
-def download_manifest(url=None, consumer=None):
-    """Task for downloading the manifest file from customer portal.
+def download_manifest(url=None, consumer=None, pool_id=None, count=None):
+    """Task for downloading the manifest file from customer portal,
+    also attach subscription to it if requires.
```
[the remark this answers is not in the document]
vijay8451: @jhutar to reduce the code duplication (all most same)  code I just enhance it.

If you still feel good to split this then I am fine with that too ?

automation_tools/__init__.py
vijay8451: Good idea ... let me just recall the validate again .

automation_tools/__init__.py
vijay8451: Move the from list to tuple in return value

vijay8451: @jhutar I used datacollection type list  because :
- I found it simple to use for scenario where len is helping to validate.
- True/False are using based on condition and helping to pass/fail assert too.

I agree with you it's not added in docstring and I could add it now if you are fine with it ?

automation_tools/__init__.py
jhutar: Thank you!

automation_tools/__init__.py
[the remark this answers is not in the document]
jhutar: I have a slight preference for that - this way download is just download. If there would be some code duplication, can you just create `_something` helper function or something?

automation_tools/__init__.py
jhutar: Yes, if you could add it, that would help. Thank you!

vijay8451: Done.

automation_tools/__init__.py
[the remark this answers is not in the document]
vijay8451: done

vijay8451:
@jhutar , addressed the comments , please review.

automation_tools/__init__.py
```diff
@@ -3093,35 +3093,97 @@ def download_manifest(url=None, consumer=None):
...
+    :return: list with True/False boolean
     """
     exp_subs_file = os.environ.get('EXP_SUBS_FILE', None)
     with open(exp_subs_file, 'r') as fp:
-        exp_subs = set([r.strip() for r in fp.readlines() if not r.strip().startswith('#')])
-    print("expected subscriptions in %s are %s" % (manifest_file, ' & '.join(exp_subs)))
+        exp_subs = {}
+        exp_subs_details = [r.strip() for r in fp.readlines() if not r.strip().startswith('#')]
+        for name in exp_subs_details:
+            exp_subs[name.split(';')[0]] = [name.split(';')[1], name.split(';')[2]]
```
JacobCallahan: alternatively:
```python
exp_subs_details = [r.strip().split(';') for r in fp.readlines() if not r.strip().startswith('#')]
for name_list in exp_subs_details:
    exp_subs[name_list[0]] = name_list[1:]
```

automation_tools/__init__.py
```diff
@@ -3093,35 +3093,97 @@ def download_manifest(url=None, consumer=None):
...
             sub_name_next = False
     print("current subscriptions in %s are %s" % (manifest_file, ' & '.join(current_subs)))
-    if exp_subs == current_subs:
-        return True
+
+    response = [True]
+    for sub in exp_subs.keys():
+        if sub in current_subs:
+            continue
+        else:
```
JacobCallahan: this could more simply be:
```python
if sub not in current_subs:
```
No need of the else or continue

automation_tools/__init__.py
```diff
@@ -3093,35 +3093,97 @@ def download_manifest(url=None, consumer=None):
...
+
+    :param url: Subscription Manager URL
+    :param consumer: A consumer hash to be used for getting the manifest
+    :param pool_id: A pool ID of subscription
+    :param count: A quantity of subscription requires to attach
+    :returns: boolean True/False
+    """
+    user = os.environ.get('RHN_USERNAME')
+    password = os.environ.get('RHN_PASSWORD')
+    string = u'{0}:{1}'.format(user, password)
```
JacobCallahan: can you use a better variable name here? string is non-descriptive.

automation_tools/__init__.py
```diff
@@ -3093,35 +3093,97 @@ def download_manifest(url=None, consumer=None):
...
+
+    if pool_id and count:
+        command = ('curl -sk -X POST -H "Authorization:Basic {0}"'
+                   ' "{1}/subscription/consumers/{2}/entitlements?pool={3}&quantity={4}"').format(
+            base64string.decode('utf-8'), url, consumer, pool_id, count)
+        response = run(command)
+        if 'updated' in response:
+            return True
+        else:
+            return False
```
JacobCallahan: more simply:
```python
return 'updated' in response
```

JacobCallahan:
@vijay8451 can you can your string formatting from %s to .format? that would make it consistent with the robottelo style guidelines.

vijay8451:
@JacobCallahan  comments are addressed in recent push , please review.

san7ket:
After Final Ack, I would like the final reviewer to ping me, so I can test this once before merge

vijay8451:
ping @JacobCallahan

JacobCallahan:
ACK. merge pending @san7ket's test

san7ket:
attaching failed /2264/console , Please fix

san7ket:
Failed to to gitlab MR, Merged the MR, re-triggering

san7ket:
2265/console Passed , Merging

PR #793, opened 2019-08-06
Fix versionlock prevents satellite rpms to install (BZ #1738199)

lpramuk:
Instead of running ```satellite-installer --no-lock-package-versions``` (permanent change) I chose to run much faster ```foreman-maintain packages unlock``` (effective till next installer run)

This command is also supposed to be run by customers when performing some exceptional package install. Satellite package such as ```foreman-discovery-image``` shouldn't be some exceptional.

That's what https://bugzilla.redhat.com/show_bug.cgi?id=1738199 is about. The issue is that the new locking mechanism assumes that all packages are either installed by default or handled by installer. That is not true, the most important example is foreman-discovery-image

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1019,11 +1019,6 @@ def setup_foreman_discovery(sat_version):
 
     :param str sat_version: contains Satellite version e.g. 6.3
     """
-    packages = (
-        'tfm-rubygem-foreman_discovery',
-        'rubygem-smart_proxy_discovery',
-        'tfm-rubygem-hammer_cli_foreman_discovery',
-    )
     admin_password = os.environ.get('ADMIN_PASSWORD', 'changeme')
 
     if sat_version == 'upstream-nightly':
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1019,11 +1019,6 @@ def setup_foreman_discovery(sat_version):
 
     :param str sat_version: contains Satellite version e.g. 6.3
     """
-    packages = (
-        'tfm-rubygem-foreman_discovery',
```
lpramuk: couldn't resist to deliver some cleanup along, sorry ;-)

automation_tools/__init__.py
```diff
@@ -1032,7 +1027,9 @@ def setup_foreman_discovery(sat_version):
         run('wget -nv -O- {0} | tar x --overwrite -C /var/lib/tftpboot/boot'.format(image_url))
     else:
         # Since 6.3, installer should install all required packages except FDI
-        run('rpm -q {0}'.format(' '.join(packages)))
+        if sat_version in ('6.5', '6.6'):
```
ntkathole: We don't need to unlock packages for 6.5.

automation_tools/__init__.py
```diff
@@ -1032,7 +1027,9 @@ def setup_foreman_discovery(sat_version):
...
+            # Check BZ 1738199 for final solution (versionlock prevents satellite rpms to install)
+            run('foreman-maintain packages unlock')
```
ntkathole: Another question is Should we lock it again?

san7ket: I think we should, but I pretty sure, we do some rpm installation in tests :(, virtwho comes to mind

lpramuk: it will lock next installer run which happen afterwards.
Anyway, PR solves it still better than now when we unlock permanently

automation_tools/__init__.py
lpramuk: I checked [6.3, 6.4, 6.5, 6.6] simply by running ```foreman-maintain packages unlock```
and only on [6.5, 6.6] the RC = 0 

But yeah the truth is that on 6.5 it actually does nothing. Going to amend then :)

sghai:
ACK.. I think we still might need to some tweaks to add full support in automation for locking and unclocking. For now I need the changes to trigger automation. So I'm merging the PR

PR #798, opened 2019-08-14
Satellite Maintain repo related changes

### LINKED ISSUE #796: Maintain repo installed via repo file even though its available via AK
For some reason why we use a repo file for the maintain file anymore, even though its included as part of the dogfood activation key/content view

https://github.com/SatelliteQE/automation-tools/blob/ff99b72bc2b6818019ee06eb107b450c46cb9190/automation_tools/__init__.py#L2239-L2243

I am going to work on fixing this if no one speaks up.

### LINKED ISSUE #797: satellite-maintenance-beta is not used when selecting "beta" distribution.
If we select "BETA" distribution in the installer job, it enables the regular satellite-maintaince repo and not the beta one. need to fix this.

pgagne:
Fix #796 Removed creation of satellite-maintain repo since it is provided as part of activation keys, repofiles, etc already.
Fix #797. When "BETA" is set (during CDN install), use beta satellite-maintenance

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2236,11 +2236,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         puppet4_repo = os.environ.get('PUPPET4_REPO')
         if puppet4_repo:
             execute(create_custom_repos, puppet4_repo=puppet4_repo, host=host)
-    # Sat6.4 and above: enable internal foreman-maintain repo for installation
-    if sat_version != '6.3' and not distribution.endswith('cdn'):
-        maintain_repo = os.environ.get('MAINTAIN_REPO')
-        if maintain_repo:
-            execute(create_custom_repos, maintain_repo=maintain_repo, host=host)
+
     # execute returns a dictionary mapping host strings to the given task's
     # return value
     installer_options.update(execute(
--- a/automation_tools/repository.py
+++ b/automation_tools/repository.py
@@ -162,11 +162,10 @@ def enable_satellite_repos(cdn=False, beta=False, disable_enabled=True,
         ]
         if sat_version in ['6.4', '6.5']:
             repos.append('rhel-{0}-server-ansible-2.6-rpms')
-            repos.append('rhel-{0}-server-satellite-maintenance-6-rpms')
         if sat_version == '6.6':
             # Ansible '2.8' was not released yet, so using '2' repo
             repos.append('rhel-{0}-server-ansible-2-rpms')
-            repos.append('rhel-{0}-server-satellite-maintenance-6-rpms')
+
     if beta:
         repos.append('rhel-server-{0}-satellite-6-beta-rpms')
     elif cdn:
@@ -174,6 +173,12 @@ def enable_satellite_repos(cdn=False, beta=False, disable_enabled=True,
         # Sat6.3: enable *cdn* puppet4 repo to perform fresh p4 install
         if sat_version == '6.3' and puppet4 == 'yes':
             repos.append('rhel-{0}-server-satellite-{1}-puppet4-rpms')
+
+    if sat_version in ['6.4', '6.5', '6.6']:
+        if beta:
+            repos.append('rhel-{0}-server-satellite-maintenance-6-beta-rpms')
+        else:
+            repos.append('rhel-{0}-server-satellite-maintenance-6-rpms')
 
     enable_repos(*[repo.format(os_version, sat_version) for repo in repos])
     run('yum repolist')


automation_tools/__init__.py
```diff
@@ -2236,11 +2236,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         puppet4_repo = os.environ.get('PUPPET4_REPO')
         if puppet4_repo:
             execute(create_custom_repos, puppet4_repo=puppet4_repo, host=host)
-    # Sat6.4 and above: enable internal foreman-maintain repo for installation
-    if sat_version != '6.3' and not distribution.endswith('cdn'):
-        maintain_repo = os.environ.get('MAINTAIN_REPO')
-        if maintain_repo:
-            execute(create_custom_repos, maintain_repo=maintain_repo, host=host)
```
ntkathole: what if distribution is internal/ internal repo profile? how maintain repo will be added?

automation_tools/repository.py
```diff
@@ -175,6 +174,12 @@ def enable_satellite_repos(cdn=False, beta=False, disable_enabled=True,
         if sat_version == '6.3' and puppet4 == 'yes':
             repos.append('rhel-{0}-server-satellite-{1}-puppet4-rpms')
 
+    if sat_version in ['6.4', '6.5', '6.6']:
```
ntkathole: we can remove this if condition as maintain is required for all sat versions.

automation_tools/__init__.py
pgagne: for INTERNAL REPOFILE, the repo file we get has an entry for the maintenance repo. 

For INTERNAL, I probably could add this code back in, in the `downstream_install` method

automation_tools/repository.py
pgagne: Thanks will do.

automation_tools/__init__.py
pgagne: @ntkathole fixed this by moving it to `downstream_install` method and adding the options for a MAINTAIN_BASE_URL to installer job see also https://github.com/SatelliteQE/robottelo-ci/pull/1570

pgagne:
@ntkathole anything else?

automation_tools/__init__.py
```diff
@@ -1891,6 +1896,9 @@ def downstream_install(admin_password=None, run_katello_installer=True):
         remote_path='/etc/yum.repos.d/satellite.repo')
     satellite_repo.close()
 
+    maintain_url = os.environ.get('MAINTAIN_BASE_URL')
```
ntkathole: duplicate code. already assigned maintain_url at 1884 line

pgagne: removed this

automation_tools/repository.py
```diff
@@ -162,11 +162,10 @@ def enable_satellite_repos(cdn=False, beta=False, disable_enabled=True,
...
         if sat_version in ['6.4', '6.5']:
             repos.append('rhel-{0}-server-ansible-2.6-rpms')
-            repos.append('rhel-{0}-server-satellite-maintenance-6-rpms')
         if sat_version == '6.6':
             # Ansible '2.8' was not released yet, so using '2' repo
             repos.append('rhel-{0}-server-ansible-2-rpms')
-            repos.append('rhel-{0}-server-satellite-maintenance-6-rpms')
+
     if beta:
         repos.append('rhel-server-{0}-satellite-6-beta-rpms')
```
san7ket: repos.append('rhel-{0}-server-satellite-maintenance-6-beta-rpms') 
Why not here??

automation_tools/repository.py
```diff
@@ -175,6 +174,11 @@ def enable_satellite_repos(cdn=False, beta=False, disable_enabled=True,
         if sat_version == '6.3' and puppet4 == 'yes':
             repos.append('rhel-{0}-server-satellite-{1}-puppet4-rpms')
```
san7ket: else: 
      repos.append('rhel-{0}-server-satellite-maintenance-6-rpms')

pgagne: @san7ket I was a little confused with what you meant here, but fixed it to what i think you meant.

automation_tools/repository.py
pgagne: @san7ket I was a little confused with what you meant here, but fixed it to what i think you meant.

automation_tools/__init__.py
lpramuk: @pgagne what about not including in downstream_install and let it be where it is now (in product_install)
and you would just change if clause to: 
```
if sat_version != '6.3' and distribution.endswith('downstream'):
```
I don like introducing new env var to downstream_install() 
I have a plan :-) to unify all ```*_install() ``` into one or two common ways of install

lpramuk: Since this PR only attempts to narrow the initial condition that is way too broad:
```
if sat_version != '6.3' and not distribution.endswith('cdn'):
```
to 
```
if sat_version != '6.3' and distribution.endswith('downstream'):
```
Why such a rewrite if you only want to narrow down one simple condition ? ;-)

pgagne:
@lpramuk any update on this.

lpramuk:
@san7ket @ntkathole 
This PR not only fixes creation of maintain repo when not needed (where my nacking comment https://github.com/SatelliteQE/automation-tools/pull/798#discussion_r317141389 is applicable)
But this PR also brings possibility to use custom URL for foreman-maintain repo as we can do with satellite repo. 

So further discussion should go this way, whether is sane to have such option (custom url for maintain repo) and to its implementation details...

automation_tools/__init__.py
```diff
@@ -1881,6 +1883,11 @@ def downstream_install(admin_password=None, run_katello_installer=True):
         print('The BASE_URL environment variable should be defined')
         sys.exit(1)
 
+    maintain_url = os.environ.get('MAINTAIN_BASE_URL')
```
ntkathole: @pgagne I think we don't need to introduce new variable here. We can simply overwrite MAINTAIN_REPO variable. See my comment at https://github.com/SatelliteQE/robottelo-ci/pull/1570/files#r320761803

Issue #800, opened 2019-09-06
Fix automation to use foreman-maintain to install packages instead of using yum.


jameerpathan111:
As per recent changes in version-locking feature of foreman-maintain, user will only be able to install/update packages using foreman-maintain.
If user wants to install/update packages freely(using yum) then he needs to either disable
packages-locking or add the name of package to whitelist file, but this approach is not supported.

Because of this change we need to make sure that instead of yum we use foreman-maintain to install packages.
Reference:- [https://github.com/theforeman/foreman_maintain/pull/278](https://github.com/theforeman/foreman_maintain/pull/278)

lpramuk:
We can run every yum command with 
```--disable-plugin=versionlock``` and yum works just fine

PR #802, opened 2019-09-13
Fix version locking package installation

ntkathole:
Issue #800

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -298,16 +298,11 @@ def setup_default_docker():
         )
 
     # Install ``Docker`` package.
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -820,7 +815,7 @@ def setup_python_code_coverage():
     if not epel_present:
         run('yum -y install https://dl.fedoraproject.org/pub/epel/'
```
lpramuk: direct package install will work I suppose

ntkathole: Yes it works!

san7ket:
We can look into improving this, For now, merging

Issue #799, opened 2019-08-19
Move away from Env Vars and use Paramerters in repo


san7ket:
In the repo , we rely on many os.enviorn() Variables to be present. 
This gets the work done, but ties us heavily with the environment. If there are changes in Infra or environment, these variables may not be present, making our automation less effective.

This issue would be a tracker for improving,

san7ket:
https://github.com/SatelliteQE/automation-tools/pull/790

PR #803, opened 2019-10-07
Enhanced manifest refresh

vijay8451:
This PR contains:
- Refresh-manifest : Remove all subscriptions from Manifest and add only required ones.
- Test result: `/satellite6-manifest-downloader/25`

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -3173,6 +3173,67 @@ def attach_subscription(url=None, consumer=None, pool_id=None, count=None):
         raise ValueError("Pass values under pool_id and count")
 
 
+def refresh_manifest(url=None, consumer=None):
+    """Task to remove all subscriptions from manifest and attach required ones.
+
+    The following environment variables affect this command:
+
+    SM_URL
+      Subscription Manager URL (e.g. 'https://subscription.rhsm.redhat.com')
+    CONSUMER
+        A consumer hash to be used for getting the manifest
+    RHN_USERNAME
+        Red Hat Network username
+    RHN_PASSWORD
+        Red Hat Network password
+
+    :param url: Subscription Manager URL
+    :param consumer: A consumer hash to be used for getting the manifest
+    :returns: boolean True/False
+    """
+    user = os.environ.get('RHN_USERNAME')
+    password = os.environ.get('RHN_PASSWORD')
+    auth_details = u'{0}:{1}'.format(user, password)
+
+    if isinstance(auth_details, str):  # py3
+        bytestring = bytes('{0}:{1}'.format(user, password), 'utf-8')
+    else:  # py2
+        bytestring = bytes('{0}:{1}'.format(user, password))
+
+    base64string = base64.encodestring(bytestring).strip()
+
+    if url is None:
+        url = os.environ.get('SM_URL')
+    if consumer is None:
+        consumer = os.environ.get('CONSUMER')
+
+    command = ('curl -sk -X DELETE -H "Authorization:Basic {0}"'
+               ' "{1}/subscription/consumers/{2}/entitlements"').format(
+        base64string.decode('utf-8'), url, consumer)
+
+    print("Cleaning up all subscriptions from Manifest!")
+    response = run(command)
+    result = 'deletedRecords' in response
+
+    if result:
+        exp_subs_file = os.environ.get('EXP_SUBS_FILE', None)
+        with open(exp_subs_file, 'r') as fp:
+            exp_subs = {}
+            exp_subs_details = [r.strip().split(';') for r in fp.readlines() if not r.strip(
+            ).startswith('#')]
+            for name_list in exp_subs_details:
+                exp_subs[name_list[0]] = name_list[1:]
+
+            for sub in exp_subs.keys():
+                print("Attaching subscription {}".format(sub))
+                attach = attach_subscription(pool_id=exp_subs[sub][0], count=exp_subs[sub][1])
+                if attach:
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -3173,6 +3173,67 @@ def attach_subscription(url=None, consumer=None, pool_id=None, count=None):
...
+    RHN_USERNAME
+        Red Hat Network username
+    RHN_PASSWORD
+        Red Hat Network password
+
+    :param url: Subscription Manager URL
+    :param consumer: A consumer hash to be used for getting the manifest
+    :returns: boolean True/False
+    """
+    user = os.environ.get('RHN_USERNAME')
```
ntkathole: please avoid using environment variables. We are moving away from using them https://github.com/SatelliteQE/automation-tools/issues/799

vijay8451: Okay I was not aware about that..
As this requires to remove from other functions as well hence would be good to handle separately or under #799

san7ket: I think it would be a easy fix here itself, just add these to parameter, and in robottelo-ci call it with those parameter. Those are already in the job. 
Not much work involved in the job as job is already a pipeline job if I am not wrong. 

Also, Is this a urgent PR?

vijay8451: thanks .. removed the env vars deps now in latest commit
Also yeah it's not a critical PR but would be great to merge asap as avoid further noises around manifest download.

vijay8451:
added 'DND' as few changes are pending at Job end @san7ket  is working on that now.

automation_tools/__init__.py
```diff
@@ -3173,15 +3173,65 @@ def attach_subscription(url=None, consumer=None, pool_id=None, count=None):
...
+    auth_details = u'{0}:{1}'.format(user, password)
+
+    if isinstance(auth_details, str):  # py3
+        bytestring = bytes('{0}:{1}'.format(user, password), 'utf-8')
+    else:  # py2
+        bytestring = bytes('{0}:{1}'.format(user, password))
+
+    base64string = base64.encodestring(bytestring).strip()
+
+    command = ('curl -sk -X DELETE -H "Authorization:Basic {0}"'
```
rplevka: why not use python `requests` module for this?

automation_tools/__init__.py
```diff
@@ -3173,15 +3173,65 @@ def attach_subscription(url=None, consumer=None, pool_id=None, count=None):
...
+               ' "{1}/subscription/consumers/{2}/entitlements"').format(
+        base64string.decode('utf-8'), url, consumer)
```
rplevka: why not use python `requests` module for this?

automation_tools/__init__.py
```diff
@@ -3173,6 +3173,67 @@ def attach_subscription(url=None, consumer=None, pool_id=None, count=None):
...
+
+    print("Cleaning up all subscriptions from Manifest!")
+    response = run(command)
+    result = 'deletedRecords' in response
+
+    if result:
+        exp_subs_file = os.environ.get('EXP_SUBS_FILE', None)
```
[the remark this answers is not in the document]
vijay8451: ahh .. nice catch ... it's pushed now .

automation_tools/__init__.py
vijay8451: It was followed  from existing one :) : https://github.com/SatelliteQE/automation-tools/blob/master/automation_tools/__init__.py#L3167

automation_tools/__init__.py
```diff
@@ -3173,15 +3173,68 @@ def attach_subscription(url=None, consumer=None, pool_id=None, count=None):
...
+    :param url: Subscription Manager URL
+    :param consumer: A consumer hash to be used for getting the manifest
+    :param user: Red Hat Network username
+    :param password: Red Hat Network password
+    :param exp_subs_file: Expected subscription file
     :param manifest_file: Specify the manifest file path.
     """
     if manifest_file is None:
         manifest_file = download_manifest()
         if os.environ.get('EXP_SUBS_FILE', None) is not None:
```
san7ket: We are still checking the environ variable here

vijay8451: @san7ket I have only replaced the the env vars which are part of new change ... however the change which you mention requires a different PR to replace every env vars from  `automation_tools/__init__.py`

san7ket: See my PR https://github.com/SatelliteQE/robottelo-ci/pull/1633
This is now passed as a parameter , So the function would not get the env variable. 
Just check the parameter here, 
`if exp_subs_file is not None:`

automation_tools/__init__.py
rplevka: Yeah, i realized. I'm just thinking whether it wouldn't be better to use requests straight away

automation_tools/__init__.py
vijay8451: ahh .. Thanks ..
its there now .

automation_tools/__init__.py
vijay8451: I like the idea and would handle here https://github.com/SatelliteQE/automation-tools/issues/807 if that is fine with you.

vijay8451: @rplevka ping :)

vijay8451:
Can we merge this and depended PR if no pending on these ?

san7ket:
@vijay8451 MR is merged, pending on  SatelliteQE/robottelo-ci#1633 I will try to get some review today. If it still pending untill EOD , I will merge and do a test run.

PR #808, opened 2019-10-23
Update accepted Sat GA versions

JacobCallahan:
Now accepts up to 6.6

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2157,10 +2157,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
 
     if (
         distribution == 'satellite6-cdn' and
-        sat_version not in ('6.3', '6.4', '6.5')
+        sat_version not in ('6.3', '6.4', '6.5', '6.6')
     ):
         raise ValueError(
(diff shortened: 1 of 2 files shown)


pgagne:
@JacobCallahan I think we also need to take a look at the version of ansible we install during the cdn install
https://github.com/SatelliteQE/automation-tools/blob/21ad57484202257f6a486f7d7f94363b9461eca2/automation_tools/repository.py#L165-L167

`rhel-7-server-ansible-2.8-rpms` is now avaliable for rhel7 so i think we should use it, it is what is recommended in the install docs.

JacobCallahan:
@pgagne that is a separate issue that someone can take up. This is targetted just for getting GA builds going.

pgagne:
@JacobCallahan I disagree, all I saying is we need to make sure that the automation matches what the docs say:
https://access.redhat.com/documentation/en-us/red_hat_satellite/6.6/html/installing_satellite_server_from_a_connected_network/installing_satellite_server#configuring-repositories_satellite

Can you change the ansible repo to `rhel-7-server-ansible-2.8-rpms` as part of this PR.

JacobCallahan:
@pgagne that's a bit of scope-creep, but updated.

PR #809, opened 2019-11-06
Deploy baseimage with suffix '-base' to use it directly

lpramuk:
Let's deploy baseimage with suffix '-base' so that we can use it directly and no other action like symlinking is needed. 
Using symlink as moving target is not a good idea. In future we better change explicitly baseimage name to be more transparent about the change

It brings us these benefits:
- older images (of the same dot release) can still be used
- snapshots (VM instances) based of older images remain viable (yay)


san7ket:
ACK

san7ket:
Pending , One thing , Can you clean up the hypervisors of the remaining copies of base images

PR #815, opened 2019-11-18
Fix package_install() failing when run in pre-install time

lpramuk:
E.g. `setup_firewall` is run before foreman-maintain cmd is present


lpramuk:
Fixed small issue: added `.succeeded`

PR #816, opened 2019-11-18
Avoid using "foreman-maintain packages install" for some "safe" packages

lpramuk:
This should speedup satellite installation avoiding un-neccessary installer runs

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -286,23 +286,14 @@ def setup_default_docker():
     os_version = distro_info()[1]
 
     # Enable required repository
-    if os_version >= 7:
-        run(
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -286,23 +286,14 @@ def setup_default_docker():
...
-    if os_version >= 7:
-        run(
-            'subscription-manager repos --enable "rhel-{0}-server-extras-rpms"'
-            .format(os_version)
-        )
-    else:
-        run(
-            'yum -y localinstall '
-            'http://mirror.pnl.gov/epel/6/x86_64/epel-release-6-8.noarch.rpm'
-        )
```
lpramuk: Sorry couldn't resist to do RHEL6 clean-up here

PR #818, opened 2019-11-26
modify katello_service command for sat_version > 6.7.0

rplevka:
katello-nightly got rid of the `katello-service` alias, we now need to use `foreman-maintain service` directly

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2286,7 +2286,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
     execute(setup_alternate_capsule_ports, host=host)
 
     execute(setup_default_docker, host=host)
-    execute(katello_service, 'restart', host=host)
+    execute(katello_service, 'restart', sat_version=sat_version, host=host)
     # if we have ssh key to libvirt machine we can setup access to it
     if os.environ.get('LIBVIRT_KEY_URL') is not None:
         execute(setup_libvirt_key, host=host)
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2942,18 +2942,19 @@ def katello_installer(debug=False, distribution=None, verbose=True,
...
     """Run katello-service
 
     :param str action: One of stop, start, restart, status.
     :param list exclude: A list of services to skip
 
     """
-    if exclude is None:
-        exclude = ''
+    exclude = '--exclude {0}'.format(','.join(exclude)) if exclude else ''
+    if (sat_version.split('.')) > ('6', '7'):  # sat_version > '6.7'
```
ntkathole: Not sure, how it will compare upstream-nightly?

ntkathole:
@rplevka also shall we just do it for upstream-nightly as of now?

automation_tools/__init__.py
rplevka: It sorts it correctly (i just had an error in the statement):
```python

In [26]: tuple('foreman_nightly'.split('.')) > ('6', '7')
Out[26]: True

In [27]: tuple('aaaa'.split('.')) > ('6', '7')
Out[27]: True
```

rplevka:
> @rplevka also shall we just do it for upstream-nightly as of now not sat-versions?
The `foreman-maintain service` method is actually already present in 6.6 and i think it should be preferred. The alias can stop working any time, I expect it to be gone with 6.8, thus i set the statement to use `> 6.7`

ntkathole:
ignore above comment, behaviour changed in recent versions

PR #819, opened 2019-12-19
enable remote-execution-cockpit feature

rplevka:
this should enable the cockpit webconsole feature in versions 6.7+
and copy the foreman-proxy public rsa key to `/root/.ssh/authorized_keys` to enable REX and cockpit access to the satellite host itself.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1091,6 +1091,13 @@ def setup_libvirt_key():
         )
 
 
+def setup_local_rex_key():
+    """Puts the foreman-proxy public RSA key to authorized_keys of root user
+    This allows performing REX jobs (and accessing cockpit console) on the sat host itself
+    """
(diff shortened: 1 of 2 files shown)


automation_tools/__init__.py
```diff
@@ -2244,6 +2251,10 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         installer_options.update({
             'foreman-proxy-plugin-pulp-puppet-content-dir': '/etc/puppetlabs/code/environments'})
 
+    # enable cockpit feature
+    if (sat_version not in ('6.3', '6.4', '6.5', '6.6')):
+        installer_options.update({'enable-foreman-plugin-remote-execution-cockpit': 'true'})
```
ntkathole: @rplevka for upstream-nightly, this condition become true and caused failure ERROR: too many arguments.

It works fine if passed `--enable-foreman-plugin-remote-execution-cockpit` instead of `--enable-foreman-plugin-remote-execution-cockpit="true"`. 

I see same behaviour for satellite 6.7. Is it expected or bug?

PR #822, opened 2020-01-06
fixing cockpit parameter value

rplevka:
Fixed the passed value (should be a flag, not a key=value param)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2257,7 +2257,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
 
     # enable cockpit feature
     if (sat_version not in ('6.3', '6.4', '6.5', '6.6')):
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2257,7 +2257,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
 
     # enable cockpit feature
     if (sat_version not in ('6.3', '6.4', '6.5', '6.6')):
```
ntkathole: need to fix this condition as well, for upstream-nightly we already have added` enable-foreman-plugin-remote-execution-cockpit` option in `def katello_installer.`

rplevka: > need to fix this condition as well, for upstream-nightly we already have added` enable-foreman-plugin-remote-execution-cockpit` option in `def katello_installer.`

i removed the line from the `katello_installer` method, would that work for  you?

PR #823, opened 2020-01-06
Extract vm_create() out of product_install()

lpramuk:
This is done on behalf of preparation for future snap-guest replacement (linchpin?)
I extracted vm_create() out of product_install() so that we have now 2 fab commands instead of 1:
```
fab -D -H root@${PROVISIONING_HOST} vm_create
fab -D -H root@${SERVER_HOSTNAME} product_install:...
```
In future we can easily replace the first command with whatever we want (while still maintaining automation-tools to install satellite)
```
linchpin ...... (or any other tools of our choice)
fab -D -H root@${SERVER_HOSTNAME} product_install:...
```

Depends on https://github.com/SatelliteQE/robottelo-ci/pull/1723

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -1098,9 +1098,13 @@ def setup_local_rex_key():
     run('cat ~foreman-proxy/.ssh/id_rsa_foreman_proxy.pub >> /root/.ssh/authorized_keys')
 
 
-def vm_create():
-    """Task to create a VM using snap-guest based on a ``SOURCE_IMAGE`` base
-    image.
+def vm_create(target_image=None, source_image=None, bridge=None, network=None):
+    """Task to create a VM using snap-guest based on a `source_image` base image.
+
+    :param str target_image: target image name (VM name)
+    :param str source_image: source image name (base image name)
+    :param str bridge: isolated VLAN bridge (static IP), if ommited br0 is used (engops DHCP IP)
+    :param str network: the name of NATed network with static IP adressing. Typically "foreman"
 
     Expects the following environment variables:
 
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -1098,9 +1098,13 @@ def setup_local_rex_key():
     run('cat ~foreman-proxy/.ssh/id_rsa_foreman_proxy.pub >> /root/.ssh/authorized_keys')
 
 
-def vm_create():
-    """Task to create a VM using snap-guest based on a ``SOURCE_IMAGE`` base
-    image.
+def vm_create(target_image=None, source_image=None, bridge=None, network=None):
```
lpramuk: at least I introduced params which vary (constant env vars remain env vars)

automation_tools/__init__.py
```diff
@@ -2055,8 +2034,7 @@ def iso_install(
         return installer_options
 
 
-def product_install(distribution, create_vm=False, certificate_url=None,
-                    selinux_mode=None, sat_version=None,
+def product_install(distribution, certificate_url=None, selinux_mode=None, sat_version=None,
```
lpramuk: Removed `create_vm` param as product_install() is no longer handling VM creation

automation_tools/__init__.py
```diff
@@ -2320,7 +2278,7 @@ def product_install(distribution, create_vm=False, certificate_url=None,
         execute(setup_bfa_prevention, host=host)
     execute(fix_qdrouterd_listen_to_ipv6, host=host)
 
-    if create_vm and 'base' in target_image:
+    if os.environ.get('PYTHON_CODE_COVERAGE') == 'true':
```
lpramuk: Cannot use removed `create_vm` param so I introduced new job param `PYTHON_CODE_COVERAGE` similar to existing `RUBY_CODE_COVERAGE`

automation_tools/__init__.py
```diff
@@ -2396,10 +2354,14 @@ def fix_hostname(entry_domain=None, host_ip=None):
             .format(host_ip, entry_domain, host))
     else:
         # Required for fixing the hostname when using satellite-installer
-        ip_addr = run(r"ping -c1 $(hostname) | awk -F\( '/icmp_seq/{print$2}' "
-                      r"| awk -F\) '{print$1}'")
-        run('echo "{0} $(hostname) $(hostname -s)" >> /etc/hosts'
-            .format(ip_addr))
+        for ip_ver in (4, 6):
```
lpramuk: `fix_hostname()`:
- now handles also IPv6 fix if IPv6 address exists on the host
- does the fix to `/etc/hosts` only if IP is not reversibly resolvable to hostname (so no hacky fix when we have proper DNS records created)

automation_tools/__init__.py
ntkathole: can we avoid setting it as env var? pass as args ?

lpramuk: we can, I did it the same way as is `RUBY_CODE_COVERAGE` env var

was thinking about just one CODE_COVERAGE but Ruby code coverage has some issues that's why it is turned off, so couldn't merge python + ruby flag easily...

PR #825, opened 2020-01-07
Do not use execute() with explicitly set host=host

lpramuk:
fabric cmdline host is no longer `PROVISIONING_SERVER` but `SERVER_HOSTNAME` itself, so there is no need to adjust majority of `execute()` calls to be run against Satellite by using `host=host` param (as default implicit host is already Satellite)

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2122,31 +2122,28 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
 
     # if host already exists still fix hostname
     execute(fix_hostname)
-    # Use fabric host (host defined on fab commandline)
-    host = env['host']
 
     # If we are using activationkey, subscribe to dogfood server
     # otherwise subscribe to CDN
     if distribution == 'satellite6-activationkey':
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2156,50 +2153,49 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
...
 
-    # execute returns a dictionary mapping host strings to the given task's
-    # return value
-    installer_options.update(execute(
-        install_tasks[distribution],
-        host=host, run_katello_installer=False
-    )[host])
+    # execute returns a dictionary mapping host strings to the given task's return value
+    installer_options.update(
+        next(iter(execute(install_tasks[distribution], run_katello_installer=False).values()))
```
lpramuk: `next(iter(d.values()))` is used in order to get first value of dict (and dict contains always only 1 value) 
it's the same as `list(d.values())[0]`

`d.values()[0]` works only in python2

lpramuk: Before the change we got `d['host']` value but by omitting `host=host` we cannot use it, so get just first value `d[0]`

PR #826, opened 2020-04-17
Add satellite 6.7 to cdn list

jameerpathan111:
```
File "/home/[*******]/workspace/satellite6-installer/automation_tools/__init__.py", line 2111, in product_install
    "Satellite version should be in [6.3, 6.4, 6.5, 6.6]"
ValueError: Satellite version should be in [6.3, 6.4, 6.5, 6.6]
```
Satellite-installer job is failing because we haven't added Satellite 6.7 to satellite-cdn list.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2105,10 +2105,10 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
 
     if (
         distribution == 'satellite6-cdn' and
-        sat_version not in ('6.3', '6.4', '6.5', '6.6')
+        sat_version not in ('6.3', '6.4', '6.5', '6.6', '6.7')
     ):
         raise ValueError(
-            "Satellite version should be in [6.3, 6.4, 6.5, 6.6]"
+            "Satellite version should be in [6.3, 6.4, 6.5, 6.6, 6.7]"
         )
 
     if selinux_mode is None:


automation_tools/__init__.py
```diff
@@ -2201,7 +2201,7 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
     installer_options.update({'katello-enable-ostree': 'true'})
 
     # enable async ssh for rex
-    if sat_version not in ('6.3', '6.4', '6.5', '6.6'):
+    if sat_version not in ('6.3', '6.4', '6.5', '6.6', '6.7'):
```
lpramuk, line 2204: wrong

automation_tools/__init__.py
```diff
@@ -2210,7 +2210,7 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
             'foreman-proxy-plugin-pulp-puppet-content-dir': '/etc/puppetlabs/code/environments'})
 
     # enable cockpit feature
-    if (sat_version not in ('6.3', '6.4', '6.5', '6.6')):
+    if sat_version not in ('6.3', '6.4', '6.5', '6.6', '6.7'):
```
lpramuk, line 2213: wrong

lpramuk: it is meant to be 6.7+

automation_tools/__init__.py
lpramuk: it is meant to be 6.7+

jameerpathan111: you mean I have to add 6.8 as well?

lpramuk: no , read this as if sat_version is above 6.6

lpramuk: enable async ssh rex for 6.7 and 6.8 
negative condition is better as you dont have to add 6.7 then 6.8 then 6.x 
`if sat_version not in [6.3, 6.4, 6.5, 6.6]`
vs.
`if sat_version in [6.7, 6.8]`

jameerpathan111: @lpramuk I read your comment after I pushed new changes. :|

PR #828, opened 2020-04-17
Add smart sat_version comparisons

lpramuk:
- added new helper function `version()`
- added sat_version param (instead of reading env variable) in all `*_install()` tasks


automation_tools/utils.py
```diff
@@ -96,6 +96,11 @@ def run_command(cmd=None):
         run(cmd)
 
 
+def version(satver='0'):
+    satver = str(satver)
+    return tuple(map(int, ('9999' if 'nightly' in satver else satver).split('.')))
```
jameerpathan111, line 101: @lpramuk I see `upstream-nightly` instead of `nightly` in [satellite6-installer.yaml](https://github.com/SatelliteQE/robottelo-ci/blob/a1506356124395679fdb7093d995181a83903ba7/jobs/satellite6-installer.yaml#L46)

jameerpathan111: What do we use `upstream-nightly` or `nightly` everywhere?

lpramuk: There was downstream-nightly once and only for 6.2 (a.k.a pre-snap1 composes)
We could eventually `s/upstream-nightly/nightly` 

My code work for both:
```>>> satver='upstream-nightly'
>>> 'nightly' in satver
True
```

automation_tools/utils.py
```diff
@@ -96,6 +96,11 @@ def run_command(cmd=None):
         run(cmd)
 
 
+def version(satver='0'):
```
mshriver, line 99: This is functional for the intent, but points to the need for a more complex implementation to handle these versions.

Consider Robottelo's `SatVersionDependentValues` class, currently only used in host_info.py.  Its implementing some of the intended behavior of `Widgetastic.utils.VersionPick`, and needs to rely on comparing versions of Satellite.

I think we have a need for a version class that can parse the version string, and make comparisons of the components in the versions - like you've done here with the tuples.

Whether its here in automation-tools, or in robottelo itself, defining a Version class, instead of just a functions, allows for more direct comparison of the versions - without having to wrap them in a function call every time.

Check out this class in miq_version, which takes advantage of functools.total_ordering, and implements `__lt__` and `__eq__`. It parses the version string similarly to what you have done here.

Taking advantage of some of this and putting it together, instead of building these if/else trees to pick things relative to the satellite versions, we could define the dictionary of the options, and pick in a single statement, like:

```
ansible_repos = Widgetastic.utils.VersionPick(
    {
        Version.lowest(): 'rhel-{0}-server-ansible-2.6-rpms',
        '6.6': 'rhel-{0}-server-ansible-2.8-rpms'
    }
)
repos.append(ansible_repos.pick(current_sat_version_string))
```

lpramuk: I dont want to pull in widgetastic. We keep `automation-tools` rather simple. This is just tool to install Satellite rather than full fledged automation suite.

Anyways this PR brings improvement over what's been there for years already.

PR #829, opened 2020-04-27
Enable RH Inventory upload

lhellebr:


lhellebr:
The parameter name has changed apparently in 6.8. Do not merge yet.

lhellebr:
I've fixed this to work with both 6.7 and 6.8. Behavior for other versions should stay the same.

Note that this causes the installation to fail for 6.8 due to [0] but I'd prefer to push this anyway since 6.8 is not yet in automation so nothing will get broken - and the BZ can be expected to be fixed soon.

[0] https://bugzilla.redhat.com/show_bug.cgi?id=1834302

lpramuk:
Once the following error is fixed in the Product we can merge this PR:
```
 Execution of '/bin/yum -d 0 -e 0 -y install tfm-rubygem-foreman_rh_cloud' returned 1: Error: Nothing to do
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/util/execution.rb:297:in `execute'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/provider.rb:101:in `execute'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/provider/package/yum.rb:223:in `install'
/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/type/package.rb:92:in `block (3 levels) in <module:Puppet>'
```

lhellebr:
@lpramuk Added BZ open conditional

lhellebr:
@lpramuk Can we merge this?

PR #831, opened 2020-05-04
set UTC time to avoid time jumps

swadeley:
Chronyd jumps the time in the first few seconds after startup. This can interrupt yum transactions. Setting the time to UTC rather than local can reduce this.

### DIFF
--- a/misc/base_image_creation/ks_rhel7_template
+++ b/misc/base_image_creation/ks_rhel7_template
@@ -15,7 +15,7 @@ network  --bootproto=dhcp --device=eth0 --ipv6=auto --activate
 # Root password
 rootpw --iscrypted ENCRYPT_PASS 
 # System timezone
-timezone America/New_York
+timezone --utc America/New_York
 # System bootloader configuration
 bootloader --location=mbr --boot-drive=vda
 #autopart --type=lvm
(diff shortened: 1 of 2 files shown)


swadeley:
* Stop chronyd before using ntpdate (to avoid chronyd noticing the jump)
* Set clock to use UTC time to avoid time jumps
* start chronyd again
* Chronyd is enabled by default, no need to enable it

lpramuk:
Nice!
We yet need to check with what TZ are VMs being created.
So that OS and HW setting match.

lpramuk:
```
  <clock offset="utc">
    <timer name="rtc" tickpolicy="catchup"/>
    <timer name="pit" tickpolicy="delay"/>
    <timer name="hpet" present="no"/>
  </clock>
```
OK.
HW clocks ticks in UTC.

I wonder if there is option to set hwclock TZ to UTC directly by `timezone America/New_York`
instead of messing with it later in %post script

swadeley:
> Nice!
> We yet need to check with what TZ are VMs being created.
> So that OS and HW setting match.

I thought time zone setting is just off-set to system clock? The clock will not jump when you change time zone settings.

swadeley:
> And there is:
> timezone America/New_York --utc
> 
> I prefer this way instead in postscript.


I am not familiar with kickstart commands but looks like that will still solve my issue.

swadeley:
@lpramuk change made.

Thank you

lpramuk:
@swadeley please rebase your PR

swadeley:
> @swadeley please rebase your PR

Yes sir! I have done it.

swadeley:
thank you

PR #833, opened 2020-05-05
Install expect package

lhellebr:


automation_tools/__init__.py
```diff
@@ -2294,6 +2294,7 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
     if os.environ.get('HOTFIX') != 'NO_HOTFIX':
         execute(apply_hotfix)
     execute(setup_rhv_ca)
+    execute(install_expect)
```
lpramuk, line 2297: what about 
`execute(package_install('expect'))`

then you would need to import `package_install` in `fabfile.py`

lpramuk: we dont want to have so many fabric tasks for every package we are about to install.
but not a blocker for merging

lhellebr: Well, I've seen fabric tasks solely for one package installation so for the sake of consistency, I did it the same way.
Do you want me to change it? If yes, why do you think I need to import package_install? It works as it is in extra fabric task so it should also work directly, right?

lhellebr:
Required by https://github.com/SatelliteQE/robottelo/pull/7777

lhellebr:
@SatelliteQE/quality-engineers This needs 1 more ACK

PR #838, opened 2020-05-22
using katello_service helper

### LINKED ISSUE #837: replace instances of katello-service
for sat 6.8 it in foreman-maintain service
same in robottelo-ci

pondrejk:
fixes #837 
Using the existing helper to pick the right service command based on sat version


pondrejk:
@lpramuk done, thanks

PR #841, opened 2020-06-02
add debug log level options for foreman and proxy

rplevka:
Setting logging levels to DEBUG to help failure investigation.
Especially in 6.8 - where all the tracebacks are now visible only with DEBUG log level.
This is useful for the previous version too though.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2847,13 +2847,17 @@ def foreman_debug(tarball_name=None, local_path=None):
 # =============================================================================
 # Helper functions
 # =============================================================================
-def katello_installer(debug=False, distribution=None, verbose=True,
+def katello_installer(distribution=None, verbose=True,
                       sat_version='6.3', scenario=None, **kwargs):
     """Runs the installer with ``kwargs`` as command options."""
     # capsule-dns-forwarders should be repeated if setting more than one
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2860,13 +2860,16 @@ def foreman_debug(tarball_name=None, local_path=None):
...
                       sat_version='6.3', scenario=None, **kwargs):
     """Runs the installer with ``kwargs`` as command options."""
     # capsule-dns-forwarders should be repeated if setting more than one
     # value check if a list is being received and repeat the option with
     # different values
-    extra_options = []
+    extra_options = [
+        '--log-level=DEBUG',
+        '--foreman-proxy-log-level=DEBUG',
+    ]
```
lpramuk, line 2872: Please add debug for foreman `--foreman-logging-level=debug`
here debug has to be lowercase !!!

automation_tools/__init__.py
```diff
@@ -2899,7 +2902,6 @@ def katello_installer(debug=False, distribution=None, verbose=True,
     run('{0}-installer --scenario {1} {2} {3} {4} {5}'.format(
```
lpramuk, line 2902: Pls remove `{5}` as you removed one item from `format(...)`

PR #842, opened 2020-06-05
Add setup_http_proxy

jameerpathan111:
@vijay8451 @mirekdlugosz I think this is what we need to change in our automation so that we can start using Http Proxy.

Ref: 
- https://github.com/SatelliteQE/robottelo-ci/pull/1793
- https://theforeman.org/manuals/2.0/index.html#5.9HTTP(S)Proxy [For more info about what those settings exactly means]

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -444,6 +444,40 @@ def setup_default_libvirt(bridge=None, ip_address="192.168.100.1"):
     return interface
 
 
+def setup_http_proxy(proxy_info):
+    """Postinstall task to set http_proxy and content_default_http_proxy settings for
+    Satellite >= 6.7
+
+    Proxy information is passed using the PROXY_INFO environmental variable.
+    The expected format is::
+
+        PROXY_INFO=proxy://<username>:<password>@<hostname>:<port>
+
+    ``username`` and ``password`` fields can be omitted if the proxy does not
+    require authentication, for example::
+
+        PROXY_INFO=proxy://<hostname>:<port>
+    """
+    command = (
+        'hammer -u admin -p {0} settings set --name "http_proxy" --value {1}'
+    ).format(os.environ.get('ADMIN_PASSWORD', 'changeme'), proxy_info.replace('proxy:', 'http:'))
+    run(command, warn_only=True)
+    proxy_info = urlsplit(proxy_info)
+    url = 'http://{0}:{1}'.format(proxy_info.hostname, proxy_info.port)
+    command = (
+        'hammer -u admin -p {0} http-proxy create --name {1} --url {2}'
+    ).format(os.environ.get('ADMIN_PASSWORD', 'changeme'), proxy_info.hostname, url)
+    if proxy_info.username and proxy_info.password:
+        creds = ' --username {0} --password {1}'.format(proxy_info.username, proxy_info.password)
+        command = command + creds
+    run(command, warn_only=True)
+    command = (
+        'hammer -u admin -p {0} settings set --name "content_default_http_proxy" --value {1}'
+    ).format(
+        os.environ.get('ADMIN_PASSWORD', 'changeme'), proxy_info.hostname)
+    run(command, warn_only=True)
+
+
 def setup_default_subnet():
     """Postinstall task to setup default subnet within Satellite
 
@@ -2162,7 +2196,8 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
     if (sat_version not in ('6.4', '6.5', '6.6', '6.8')):
         installer_options.update({'enable-foreman-plugin-remote-execution-cockpit': None})
 
-    if os.environ.get('PROXY_INFO'):
+    proxy_info = os.environ.get('PROXY_INFO')
+    if proxy_info and version(sat_version) < version(6.7):
         # execute returns a dictionary mapping host strings to the given
         # task's return value
         installer_options.update(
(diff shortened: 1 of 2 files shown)


vijay8451:
@jameerpathan111 LGTM :+1:  ... @lpramuk Are we good to run the automation with newly proxy now ?

lhellebr:
LGTM but I'm not a Tier{1,2} reviewer...

jameerpathan111:
> @jameerpathan111 LGTM +1 is there any related test results with new changes ?

There shouldn't be need for new test result because I have only changed one variable name.

jameerpathan111:
@vijay8451 @mirekdlugosz @lpramuk I have rebased this PR. Need your ack/review.

automation_tools/__init__.py
```diff
@@ -444,6 +444,42 @@ def setup_default_libvirt(bridge=None, ip_address="192.168.100.1"):
...
+
+    ``username`` and ``password`` fields can be omitted if the proxy does not
+    require authentication, for example::
+
+        PROXY_INFO=proxy://<hostname>:<port>
+    """
+    proxy_info = os.environ.get('PROXY_INFO')
+    if proxy_info is None:
+        print('The PROXY_INFO environment variable should be defined')
+        sys.exit(1)
```
mirekdlugosz, line 464: This function is gated behind `if os.environ.get('PROXY_INFO')`. I wonder in what scenario we can call it and then have variable undefined under our feet.

automation_tools/__init__.py
```diff
@@ -444,6 +444,42 @@ def setup_default_libvirt(bridge=None, ip_address="192.168.100.1"):
...
+    command = (
+        'hammer -u admin -p {0} settings set --name "http_proxy" --value {1}'
```
mirekdlugosz, line 466: I'm not familiar with automation tools and entire workflow enough, but is username `admin` in every single case? Or should we pull it from variable as well (same as password)?

automation_tools/__init__.py
```diff
@@ -444,6 +444,42 @@ def setup_default_libvirt(bridge=None, ip_address="192.168.100.1"):
...
+    ).format(os.environ.get('ADMIN_PASSWORD', 'changeme'), proxy_info.replace('proxy:', 'http:'))
+    run(command, warn_only=True)
+    proxy_info = urlsplit(proxy_info)
+    url = 'http://{0}:{1}'.format(proxy_info.hostname, proxy_info.port)
+    creds = '--username {0} --password {1}'.format(proxy_info.username, proxy_info.password)
```
mirekdlugosz, line 471: Will these be always defined? I'm worried it will break if username or password ends up empty.

automation_tools/__init__.py
jameerpathan111: Yes, username `admin` has been used in all the tasks.

automation_tools/__init__.py
jameerpathan111: @mirekdlugosz thanks for noticing that. I have updated the PR accordingly :D

automation_tools/__init__.py
jameerpathan111: @mirekdlugosz Updated the  PR.

automation_tools/__init__.py
```diff
@@ -444,6 +444,41 @@ def setup_default_libvirt(bridge=None, ip_address="192.168.100.1"):
...
+    command = (
+        'hammer -u admin -p {0} settings set --name "http_proxy" --value {1}'
+    ).format(os.environ.get('ADMIN_PASSWORD', 'changeme'), proxy_info.replace('proxy:', 'http:'))
+    run(command, warn_only=True)
+    proxy_info = urlsplit(proxy_info)
+    url = 'http://{0}:{1}'.format(proxy_info.hostname, proxy_info.port)
+    command = (
+        'hammer -u admin -p {0} http-proxy create --name {1} --url {2}'
+    ).format(os.environ.get('ADMIN_PASSWORD', 'changeme'), proxy_info.hostname, url)
+    if proxy_info.username and proxy_info.password is not None:
```
mirekdlugosz, line 471: Didn't you mean one of
```suggestion
    if proxy_info.username is not None and proxy_info.password is not None:
```
```suggestion
    if proxy_info.username and proxy_info.password:
```
It's little weird to explicitly check against None for one variable, but for falseness for another.

jameerpathan111: Yes, Thanks. Updated PR

automation_tools/__init__.py
```diff
@@ -444,6 +444,41 @@ def setup_default_libvirt(bridge=None, ip_address="192.168.100.1"):
     return interface
 
 
+def setup_http_proxy():
```
lpramuk, line 447: can we have `proxy_info` as a param here?

automation_tools/__init__.py
```diff
@@ -2250,6 +2285,8 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
         execute(apply_hotfix)
     execute(setup_rhv_ca)
     execute(install_expect)
+    if os.environ.get('PROXY_INFO') and sat_version >= '6.7':
```
lpramuk, line 2288: Pls use new `version()` function:
```
if proxy_info and version(sat_version) >= version(6.7):
    execute(setup_http_proxy, proxy_info)
```

automation_tools/__init__.py
```diff
@@ -2162,7 +2197,7 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
     if (sat_version not in ('6.4', '6.5', '6.6', '6.8')):
         installer_options.update({'enable-foreman-plugin-remote-execution-cockpit': None})
 
-    if os.environ.get('PROXY_INFO'):
+    if os.environ.get('PROXY_INFO') and sat_version < '6.7':
```
lpramuk, line 2200: Can you pls read env var once and use new `version()` function helper?
```
proxy_info = os.environ.get('PROXY_INFO')
if proxy_info and version(sat_version) < version(6.7):
```

automation_tools/__init__.py
jameerpathan111: Thanks @lpramuk for comments. I have updated the PR accordingly.

PR #845, opened 2020-06-22
Refactor manifest tasks

lpramuk:
- moved manifest tasks to `manifest.py`
- added helper `read_subs()` for parsing expected subscriptions file (aka. manifest definition)
- added function `read_manifest()` to read what subscriptions the manifest contains
- added function `delete_subscriptions()` for deleting all subscriptions from the manifest
- refactored `validate_manifest()` so that it returns true if manifest has been modified
- refactored `relink_manifest()` so that it modifies manifest **only if needed** and downloads updated manifest **after** it was modified


lpramuk:
Tested by modifying job to run from my fork/branch:
`/job/satellite6-manifest-downloader/395/`

PR #846, opened 2020-07-07
Add workaround for BZ 1852371 (http_proxy bug)

lpramuk:
There is a SELinux bug in 6.8 which prevents http proxy connections to port 3128/tcp (`squid_port_t`) 
Only allowed port type is `http_cache_port_t`. When the BZ is fixed the `squid_port_t` type will be allowed too (as was in 6.7)

Fixes the following error in automation:
```
robottelo.cli.base.CLIReturnCodeError: CLIReturnCodeError(return_code=70, stderr='Could not enable repository:\n  Failed to open TCP connection to env-proxy-02.infra:3128 (Permission denied - connect(2) for "env-proxy-02.infra" port 3128)\n'
```

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -458,23 +458,27 @@ def setup_http_proxy(proxy_info):
 
         PROXY_INFO=proxy://<hostname>:<port>
     """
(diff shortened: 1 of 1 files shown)


jameerpathan111:
Well travis failed :|

PR #849, opened 2020-07-27
Added regex for correct discovery package selection

devendra104:
This issue comes in 6.8, where there are multiple numbers of packages with the same prefix(foreman-discovery-image) name, due to that correct foreman-discover-image package selection gets failed.


latran:
I'll ack once you addressed @jyejare's suggestion.  Either or works.

PR #850, opened 2020-08-24
disable async ssh

jameerpathan111:
As per Dev's comment on BZ, leapp feature (restarting a host via rex) doesn't require async ssh option to be enabled.
This option  was causing remote execution job to show success even when the original job failed at host end.

Refs: PR #821 
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1866689#c7


mirekdlugosz (approved):
ACK

lvrtelov (approved):
ACK. I can confirm that LEAPP feature doesn't need async on.

pondrejk (approved):

PR #853, opened 2020-09-02
Save 2 extra installer runs

lpramuk:
We needlessly run installer two times in post-install time.
Get rid of it.


mirekdlugosz (approved):
Looks good to me

JacobCallahan (approved):
ACK

PR #856, opened 2020-11-03
Work around installer bug in 6.9 presnaps

lpramuk:
Fixes:
```
[ERROR 2020-10-29T15:27:59 verbose]  Evaluation Error: Error while evaluating a Resource Statement, Class[Dhcp]: has no parameter named 'conf_dir_mode' (file: /usr/share/foreman-installer/modules/foreman_proxy/manifests/proxydhcp.pp, line: 50, column: 3) on node qe-sat69-rhel7.example.com
```

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -2156,9 +2156,15 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
 
     # enable cockpit feature
     if version(sat_version) > version(6.6):
-        # WORKAROUND https://bugzilla.redhat.com/show_bug.cgi?id=1889320 manifests only on 6.8
-        if not bz_bug_is_open(1889320) or sat_version != '6.8':
+        # WORKAROUND https://bugzilla.redhat.com/show_bug.cgi?id=1889320 manifests only on 6.8+
+        if not bz_bug_is_open(1889320) or sat_version not in ['6.8', '6.9']:
             installer_options.update({'enable-foreman-plugin-remote-execution-cockpit': None})
+
+    # WORKAROUND for 6.9 presnaps
+    if sat_version == '6.9':
+        execute(lambda: run(
+            "sed -i '/ conf_dir_mode/d' /usr/share/foreman-installer/"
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -2156,10 +2156,16 @@ def product_install(distribution, certificate_url=None, selinux_mode=None, sat_v
...
     if version(sat_version) > version(6.6):
-        # WORKAROUND https://bugzilla.redhat.com/show_bug.cgi?id=1889320 manifests only on 6.8
-        if not bz_bug_is_open(1889320) or sat_version != '6.8':
+        # WORKAROUND https://bugzilla.redhat.com/show_bug.cgi?id=1889320 manifests only on 6.8+
+        if not bz_bug_is_open(1889320) or sat_version not in ['6.8', '6.9']:
             installer_options.update({'enable-foreman-plugin-remote-execution-cockpit': None})
 
+    # WORKAROUND for 6.9 presnaps
+    if sat_version == '6.9':
+        execute(lambda: run(
```
jameerpathan111, line 2165: I think upcoming snap won't be a presnap but an actual snap, so I don't think this is still needed as per comment on 6.9-PresSnapIssues sheet.

jameerpathan111 (approved):
But no harm in adding it as well, so ACK.

jyejare (commented):
@jameerpathan111 There should be one more presnap, so we shld be good with it for now.

But the issue this will run even on actual snaps which should not cause harm to the snap 1 and so on. But not sure even that conf will exist in 2.3, @lpramuk please confirm that before merging !

ACK pending confirmation!

lpramuk:
@jyejare lets remove this prior `Snap 1.0`
I'm not going to investigate beyond.  If it were a BZ then I would investigate and noted there.

jyejare:
@lpramuk If the plan is to remove before snap 1.0, I am good with that. But lets not carry fwd that to 1.0 else our first build results of 6.9 would be broken and potentially adding to the re trigger.

lpramuk:
@jyejare I'm more concerned about continuing to work around and hiding persisting issue. 

Ok, then it means I have your ACK ?

jyejare (approved):

PR #861, opened 2021-01-28
Remove installer cockpit workarounds for BZ 1890923 and BZ 1889320

lpramuk:
both BZs are now verified


tstrych (approved):
I read BZ's and as lukas said he verified both of them :+1: 
thanks for cleanup :) 
hopefully cockpit test is going to be green again :)

PR #862, opened 2021-01-28
Preserve pip python2 compatibility for Code Coverage

lpramuk:
This PR unblocks old CI automation.

pip-21.0 no longer supports python2 but we are still required to use python2 for pulp codecoverage


rplevka (approved):

mshriver (approved):
Approved, but for this library to move forward along with robottelo, things need to be updated to py3.

lpramuk:
> Approved, but for this library to move forward along with robottelo, things need to be updated to py3.

No, it's because pulp is still running on python2, not us

PR #863, opened 2021-03-03
Subscribe satellite as the ddns package requires java package

omkarkhatavkar:
There is an update in the ddns package url, which needs the java and causing failure of the job.  see the custom cert job 186 it is fixed with this line of code change. The remaining code needs to be updated in the gitlab repo.

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -172,7 +172,7 @@ def setup_ddns(entry_domain, host_ip):
     if ddns_hash is None:
         print('The DDNS_HASH environment variable should be defined')
         sys.exit(1)
-
+    execute(subscribe, host=host_ip)
     ddns_package_url = os.environ.get('DDNS_PACKAGE_URL')
     if ddns_package_url is None:
         print('The DDNS_PACKAGE_URL environment variable should be defined')


automation_tools/__init__.py
```diff
@@ -172,7 +172,7 @@ def setup_ddns(entry_domain, host_ip):
     if ddns_hash is None:
         print('The DDNS_HASH environment variable should be defined')
         sys.exit(1)
-
+    execute(subscribe, host=host_ip)
```
devendra104, line 175: It will work, but hostname would be better instead of IP.

devendra104 (approved):
Ack.

automation_tools/__init__.py
omkarkhatavkar: I tried that earlier and failed with `AttributeError: 'NoneType' object has no attribute 'strip'`. please take a look 185 job output

jyejare (approved):
ACK pending required changes in CI.

jyejare:
@omkarkhatavkar Until you add the update the code in CI dont merge this PR, else it will fail the CI runs. Mark your PR `draft`.

omkarkhatavkar:
@jyejare no required changes in CI

jyejare:
As per discussion with @omkarkhatavkar , CI MR is already open and after reviewing and merging that PR , merging this PR with no know side effects.

PR #864, opened 2021-03-08
fixed the java issues in the custom cert job

omkarkhatavkar:
Custom Job was failing because of the below issues, this fix will take care of all of the problems. 

- Package version change for DDNS in HDN repo 
- Need for java because of the dependency with the DDNS package
- Needs the subscription with RHEL Repos for JAVA
- Need to unsubscribe with RHEL repos and need to uninstall the java version creates the conflict with satellite installation after generating DNS record

### DIFF
--- a/automation_tools/__init__.py
+++ b/automation_tools/__init__.py
@@ -172,38 +172,47 @@ def setup_ddns(entry_domain, host_ip):
     if ddns_hash is None:
         print('The DDNS_HASH environment variable should be defined')
         sys.exit(1)
-    execute(subscribe, host=host_ip)
-    ddns_package_url = os.environ.get('DDNS_PACKAGE_URL')
-    if ddns_package_url is None:
-        print('The DDNS_PACKAGE_URL environment variable should be defined')
-        sys.exit(1)
-
-    os_version = distro_info()[1]
-
-    target, domain = entry_domain.split('.', 1)
-
-    if os_version >= 7:
-        internal_cert_url = os.environ.get('internal_cert_url')
-        if internal_cert_url is None:
-            print('The internal_cert_url environment variable should be defined')
+    # subscribed, because ddns pacakge requires java
+    try:
+        execute(subscribe, host=host_ip)
+        ddns_package_url = os.environ.get('DDNS_PACKAGE_URL')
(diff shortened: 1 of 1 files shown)


automation_tools/__init__.py
```diff
@@ -204,6 +205,9 @@ def setup_ddns(entry_domain, host_ip):
         fix_hostname(entry_domain, host_ip)
         run('redhat-ddns-client enable')
         run('redhat-ddns-client')
+    # unsubscribe and removed java as it will conflict satellite installation
+    unsubscribe()
```
jyejare, line 209: Try subscription and Catch un-subscription, so that any issues in between should not skip un-subscription.

jyejare (changes requested):
ACK pending changes.

automation_tools/__init__.py
omkarkhatavkar: No thought process is that no exception handling is required here.

devendra104 (approved):
ACK

automation_tools/__init__.py
jyejare: Lets say u subscribed and then something goes wrong with lines below that like in `fix hostname` we just dont want to exit out of that function if that func is called with `warn_only=true`, coz that will skip unsubscription and hence failing later running tests.

omkarkhatavkar: @jyejare This is the pre-setup part for running the test case. We should be halting the execution in case if the problem persists here. I wanted to have hard assert here rather than soft assert. If you have hard feelings that this needs to be added as part of try-catch I will add it.

omkarkhatavkar: done !

jyejare: Just a Suggestion:
1. Catch the exception bu raise it.
2. Add finally block for unsubscribe so need not to call twice.

Example:
try:
    bla 
    bla
except:
    print('error!')
    raise Exception
finally:
    unsubscribe()

omkarkhatavkar: you are right, added finally.

jyejare (approved):

devendra104 (approved):
ACK

PR #865, opened 2021-05-06
Fix obsoleted easy_install by using wget to work around failing download

lpramuk:
Fixes provisioning job failing at `setup_python_code_coverage` task:
```
[root@sat.example.com] run: easy_install "pip<21"
[root@sat.example.com] out: Searching for pip<21
[root@sat.example.com] out: Reading https://pypi.python.org/simple/pip/
[root@sat.example.com] out: Couldn't find index page for 'pip' (maybe misspelled?)
[root@sat.example.com] out: Scanning index of all packages (this may take a while)
[root@sat.example.com] out: Reading https://pypi.python.org/simple/
[root@sat.example.com] out: No local packages or download links found for pip<21
[root@sat.example.com] out: error: Could not find suitable distribution for Requirement.parse('pip<21')
```


JacobCallahan:
Can you try    python -m ensurepip --upgrade

lpramuk:
ensurepip is provided by `python3-libs-3.6.8-18.el7.x86_64`  
this is for python2

